PR #49, opened 2026-06-16
fix(agents): bound Anthropic key resolve with timeout + single-flight

sanjulaonline:
## Problem

`resolveAnthropicKey` runs on the hot path of **every** agent request (architect, tech-lead, requirements-chat, etc.). On a cache miss it called `asdlc-api` with two latency hazards:

- **No timeout** — a hung/slow BFF would hang the agent request indefinitely (unbounded tail latency).
- **No single-flight** — concurrent misses for one org (cold start, or right after a `/cache/invalidate`) fired N parallel resolves at the BFF.

## Fix

- **Bounded timeout** via `AbortSignal.timeout`, covering both the request *and* the response-body read (a BFF that sends `200` headers then stalls the body is still a hung upstream). Configurable via `ANTHROPIC_KEY_RESOLVE_TIMEOUT_MS` (default `5000`ms), read at call time. Timeouts surface as the existing `resolver_unreachable` (502).
- **Single-flight coalescing** keyed by orgId: concurrent misses await one shared fetch. The cache read + map ops run synchronously with no intervening `await`, so at most one fetch is in flight per org. Errors and `"none"` stay uncached, so a newly configured key is still picked up immediately.

No change to existing error codes/statuses or the 5-minute cache semantics.

## Tests

Adds `agents/src/shared/anthropic-key-resolver.test.ts`:

- coalescing → exactly 1 upstream call for 6 concurrent callers
- cache hit + refetch-after-invalidate
- fast timeout on a hung upstream
- `"none"` rejects and is not cached
- non-2xx → `resolver_error`
- empty-org guard

`npm run typecheck` clean; full agents suite **64/64** passing.

### DIFF
--- a/agents/src/shared/anthropic-key-resolver.test.ts
+++ b/agents/src/shared/anthropic-key-resolver.test.ts
@@ -1,0 +1,202 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


xlight05:
Sorry, we were working on a significant rewrite, 
Closing as this is outdated now.

If you are willing to contribute more, happy to provide support, https://discord.gg/wUh34rEG8

Issue #98, opened 2026-07-07
Console: production-serve apps/console at :8091 (legacy keeps :8090 until parity, then ports flip)


hevayo:
## Problem

The platform still serves only the **legacy console** (`console-legacy/console`), while the **new console** (`apps/console` — Thunder SSO, Oxygen UI, contract-generated client) only runs as a Vite dev server against MSW mocks. Until the new console is production-served against the real stack, real usage and feedback accrue exclusively to the codebase we intend to delete, and the new console's real-API wiring goes unexercised.

## Users

All console users of the local AEP stack (developers/platform engineers running `deployments/`).

## Experience walkthrough

1. `bash scripts/start.sh` brings up the stack. **http://localhost:8090 keeps serving the legacy console** — existing developer workflows are untouched.
2. **http://localhost:8091 serves the new console** in production mode (nginx, real BFF, real Thunder OIDC via the `/callback` route).
3. On the new console: organizations, project listing/create, project overview, board/status/components all hit the real BFF through the same-origin `/aep-api-service` proxy.
4. The spec page (once #80 merges) renders its **error state** in real mode: `GET /projects/{p}/spec` is in the contract but not yet implemented in aep-api (known gap, BE issue #99). Collab degrades to solo per #86 decision 10.
5. As parity features land, the retirement checklist below shrinks. The final PR **flips the ports** (new console takes :8090), deletes `console-legacy/`, and removes the extra Thunder redirect URIs/origins.

## Scope

**In:**
- `apps/console` production packaging: `Dockerfile` + `nginx.conf` + `docker-entrypoint.sh`, copy-adapted from the legacy container (runtime DNS resolver, SSE `proxy_buffering off`, WebSocket `/collab/` block, `/api/v1/github/` passthrough). Entrypoint generates `env-config.js` with the new console's `RuntimeEnv` keys (`VITE_API_BASE_URL`, `VITE_THUNDER_URL`, `VITE_THUNDER_CLIENT_ID`, `VITE_THUNDER_SCOPES`) plus `collabWsUrl`; drops dead legacy keys (`VITE_DEV_BYPASS_AUTH`, `BILLING_API_BASE_URL`).
- `deployments/docker-compose.yml`: `console` service builds `apps/console` (repo-root build context — pnpm workspace) at **8091:3000**; `console-legacy` service keeps the old image at **8090:3000** (container name `aep-console` retained so nothing referencing it breaks).
[...]

Issue #91, opened 2026-07-06
Console SSO — Thunder login (oidc-client-ts) + header org/project switchers


hevayo:
## Problem

The new console has no login. The BFF (`services/aep-api`) already verifies Thunder JWTs on every `/api/` request and fails closed 401, so the console can only run against MSW mocks today. The seams are placeholders: the `auth401` middleware in `src/api/client.ts` just warns (`TODO(auth)`), `AppLayout` renders a hardcoded `devUser`, and the collab editor sends a fake unsigned token (`useCollabSpec.ts` `TODO(auth)`). Legacy console had working Thunder SSO (`@asgardeo/react`); the new console needs its own, on its own stack.

## Users

Developer/User — signs into the platform console and works within their organization.

## Experience walkthrough

1. Visiting any route unauthenticated redirects to Thunder's hosted sign-in; after login the user lands back on the URL they asked for (dedicated `/callback` route handles the `?code&state` exchange with a spinner).
2. The header gains a **Switchers section** between the logo and the spacer (`Header.Switchers` + `ComplexSelect`, the Oxygen UI canonical pattern; visual parity with legacy `AepLayout`):
   - **Org switcher** (always visible): current org resolved from token claims; the other orgs from `GET /organizations` are listed but **disabled** with a "switching orgs requires re-login — coming soon" hint. No cosmetic switching — legacy's navigation-only switch silently showed wrong-org labels over token-org data, and we are not reproducing that.
   - **Project switcher** (visible inside `/projects/$projectName/*`): options from `GET /projects`; switching swaps the `$projectName` param and **keeps the current sub-page** (builds→builds, spec→spec).
3. The `UserMenu` shows the signed-in name/email from ID-token claims and offers **Sign out** → Thunder RP-initiated logout (`signoutRedirect` + `post_logout_redirect_uri`) → back to console → login redirect.
4. Sessions renew silently via refresh tokens (`automaticSilentRenew`). A 401 mid-session self-heals: silent renew → retry the request once → if still failing, full `signinRedirect` preserving the current URL.
5. The collab spec editor connects with the **real access token** (async token getter on `HocuspocusProvider`, so reconnects pick up refreshed tokens).

## Scope

[...]

Issue #72, opened 2026-07-03
aep-api: prompt on create-project (spec derivation kickoff) + search on list-projects


hevayo:
BE handshake for console feature #71 (projects listing + requirement-first create), per the console development flow step 5. Proposed `packages/contracts/api/v1/openapi.yaml` changes:

## 1. `CreateProjectRequest` gains a `prompt` field

```yaml
CreateProjectRequest:
  properties:
    # ...existing name / displayName / description / deploymentPipeline...
    prompt:
      description: >-
        The user's initial requirement — what they want built. Persisted as
        the project's requirement; triggers spec derivation for the new
        project.
      type: string
```

Behavior: `POST /projects` with a `prompt` persists it as the project's initial requirement and kicks off spec derivation (the same pipeline that turns a requirement into spec/design files). Projects created without a prompt keep today's behavior. Console always sends one (ADR-0005: creation is requirement-first).

## 2. `list-projects` gains a `search` query parameter

```yaml
/projects:
  get:
    parameters:
      - in: query
        name: search
        required: false
        schema:
          type: string
        description: Case-insensitive substring match on name and displayName.
```

Behavior: filters the returned items server-side; absent/empty returns all (unchanged).

## Also confirm (no change expected)

The console will preview the to-be-created repo URL by convention (connected GitHub org + project name). Please confirm the org/login is readable from the existing `get-github-status` operation and the repo naming convention is `<org>/<projectName>` — if not, say what it is.

## Rationale

Console feature #71 makes the projects grid the landing page with server-side search, and makes "Create project" a requirement-first flow (prompt → name/repo confirmation → create → project overview). Decisions and rejected alternatives: https://github.com/wso2/labs-agentic-engineer/issues/71 (grilling-outcome comment).

FE and BE ship separate PRs referencing this issue.

hevayo:
Addition (scope amendment on #71, 2026-07-03): the projects grid now paginates incrementally (View more button). Please also add an optional `nextCursor` string to `ProjectList` — cursor for the next page, absent on the last page — so the existing `cursor` query param is followable. Page size stays a server default; no `limit` param needed. Proposed spec diff is in the FE branch alongside `prompt` and `search`.

hevayo:
Two more additions from the #71 feedback round (2026-07-04):

1. `CreateProjectRequest.repoName` (optional string) — name of the GitHub repo to create for the project's specs/source, in the connected org; defaults to the project name. The org is not client-controllable.
[...]

PR #106, opened 2026-07-07
aep-api: accept prompt + repoName on create-project (fixes 422 from the new console)

hevayo:
The new console's create flow POSTs `{name, prompt[, repoName]}`. `prompt` is in the contract (added by #91, tied to #72) but was never added to `models.CreateProjectRequest` — huma derives the request schema from the Go struct with `additionalProperties:false`, so every create from the new console failed with `422 unexpected property: body.prompt`. `repoName` (sent when the user edits the repo name in the form) was missing from both the contract and the struct. The drift was invisible because OAS autogen was disabled (b60683b) and `openapi-check` is no longer wired into any target.

## Changes

- `models.CreateProjectRequest`: add `Prompt` (accepted per contract, unconsumed until spec derivation lands — #72) and `RepoName`.
- `project_service.CreateProject`: when `RepoName` is set it is passed to `repoSvc.CreateRepo` in place of the project name — the git repo gets the override, the OC project keeps its name.
- create handler: `repoName` validated as a DNS-label slug (400, reusing `validate.Slug`) so GitHub can't silently normalize it into a repo whose actual name diverges from what we store.
- contract: `repoName` added to `CreateProjectRequest` (`prompt` was already declared).
- TDD: component-tier test (prompt+repoName → 201; bad slug → 400 naming repoName) and service-tier test (CreateRepo receives the override; default unchanged).

## Verified

- `go test ./...` green except the pre-existing environmental `internal/platform/gitfs` failure (isolated git env can't auto-detect an identity; fails identically on the base commit).
- Live on the local stack: create from the new console (:8091) with a prompt and an edited repo name → 201, BFF log `creating repository … name=gym-tracker-repo`, repo provisioned on GitHub under the overridden name, project page renders with the repo link.
- `pnpm --filter @aep/console gen && typecheck` green — generated type gains `repoName?`, typing the console's existing conditional spread; no console code change.

## Follow-up worth its own issue

Nothing guards contract↔struct drift anymore (`openapi-check` exists in `services/aep-api/Makefile` but is unwired) — this bug is exactly the failure mode it used to catch.

Refs #98, #72

### DIFF
--- a/packages/contracts/api/v1/openapi.yaml
+++ b/packages/contracts/api/v1/openapi.yaml
@@ -746,6 +746,9 @@ components:
         prompt:
           description: Optional requirement prompt that kicks off spec derivation (issue #72).
           type: string
+        repoName:
+          description: Optional Git repository name override (DNS-label slug); defaults to the project name.
+          type: string
       required:
         - name
       type: object
--- a/services/aep-api/internal/feature/component/component_ensure_test.go
+++ b/services/aep-api/internal/feature/component/component_ensure_test.go
@@ -33,7 +33,7 @@ type ensureRepoSvc struct{ repo *models.GitRepository }
 func (r ensureRepoSvc) GetRepo(context.Context, string, string) (*models.GitRepository, error) {
 	return r.repo, nil
 }
-func (ensureRepoSvc) CreateRepo(context.Context, string, string, string) (*models.GitRepository, error) {
+func (ensureRepoSvc) CreateRepo(context.Context, string, string, string, string) (*models.GitRepository, error) {
 	panic("CreateRepo not expected")
 }
 func (ensureRepoSvc) EnsureBareRepo(context.Context, string, string, string) (*models.GitRepository, error) {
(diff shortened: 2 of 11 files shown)


hevayo:
Second commit (598f898) per review feedback: a user-chosen `repoName` is now used **verbatim** — no random 3-digit suffix — and a name conflict **fails the create with 409** instead of being suffixed away.

- `gitrepo.CreateRepo` gains an explicit `repoName` param: empty keeps the derived slug+suffix behavior with conflict retries; non-empty creates the repo exactly as given, and `ErrRepoNameConflict` propagates.
- `project.CreateProject` compensates the just-created OC project away on an explicit-name conflict (retrying the same name can never succeed) and fails the create; all other repo failures remain best-effort.
- `mapProjectError` maps the conflict to `409 — a repository with this name already exists — choose another repository name`, rendered inline by the console's create form.

Verified live on the local stack: taken name → 409 alert on the form + OC project compensated in the same correlation id; fresh name → GitHub repo created with the exact name.

hevayo:
Third commit: the random suffix is gone **entirely** — the derived (project-name) repo name is also created verbatim now. `createGitHubRepoWithRetry` is deleted; **any** name conflict (user-chosen or derived) rolls the OC project back and returns the 409 so the user picks a different name.

Verified live: project named `demo-july-7999` (existing repo) with untouched repo field → inline 409 on the form, no stranded project; project `repo-exact-check` → repo created as exactly `repo-exact-check`.

PR #124, opened 2026-07-08
deploy: seed test users into the Administrators group (#123)

### LINKED ISSUE #123: aep-api: GET /projects/{projectName} returns 403 for regular users — project layout never renders
Found while verifying #117 on the deployed compose stack (aep-rewrite HEAD, real Thunder login as a seeded test user):

```
GET /api/v1/projects/store-handmade-ceramics          → 403 {"title":"Forbidden","detail":"insufficient permissions to perform this action"}
GET /api/v1/projects/store-handmade-ceramics/status   → 200
GET /api/v1/projects/store-handmade-ceramics/files    → 200
GET /api/v1/projects/store-handmade-ceramics/tags     → 200
```

Impact: the console's `ProjectLayout` blocks on `useProject` (get-project), so the entire project view — overview cards, components, everything under the layout — renders the error state for regular users. The spec route still works because it only needs status/files. It also explains why the spec header shows the raw project slug instead of the display name.

The sibling repo-backed reads all pass the tenant gate, so this looks specific to the get-project path (OpenChoreo project read / its authz mapping) rather than the JWT or org resolution. Not a regression from #117/#118 — those PRs touch neither get-project nor authz.

Repro: seeded dev cluster, `deployments/scripts/seed-test-users.sh` user (e.g. mark/admin), open http://localhost:8091/projects/<any-project>.

hevayo:
Closes #123.

## Diagnosis

The BFF forwards user JWTs to OpenChoreo for user-facing project calls, and OC's entire user-authz surface is one binding: `administrators-group-binding` (wso2-ae-oc-extensions) maps the token's `groups=Administrators` claim to the admin ClusterAuthzRole. Seeded test users had no group at all → no OC entitlement → `GET /projects/{p}` 403 ("insufficient permissions"), `GET /projects` 200-but-empty, and the console's project layout blocks on get-project, so nothing under it ever rendered. The repo-backed reads (`/status`, `/files`, `/tags`) never touch OC, which is why they all worked.

## Fix (dev seeding)

`seed-test-users.sh` now ensures every test user is a member of Thunder's `Administrators` group — `POST /groups/{id}/members/add` (the shape is from Thunder's `api/group.yaml`; `PUT /groups/{id}` silently ignores a `members` field). Idempotent: already-members are skipped. Matches the binding's stated intent ("human operators logged into the console have immediate admin rights"). Uses `jq`, like `patch-thunder-new-console.sh`.

## Verified live

Re-ran the seeder (mark skipped as already-member from probing, john/chris/emily added), fresh `john` login: token now carries `groups: ["Administrators"]`, `GET /projects/{p}` → 200, and the full project view renders — Spec/Build/Deployment cards + Components — for a regular seeded user for the first time.

## Not in scope

Real-user onboarding entitlements (what non-admin role a JIT-onboarded org member should get, e.g. a Members group → member role binding) — filed separately as a design follow-up.

### DIFF
--- a/deployments/scripts/seed-test-users.sh
+++ b/deployments/scripts/seed-test-users.sh
@@ -124,5 +124,50 @@ for row in "${USERS[@]}"; do
     esac
 done
 
+# 4. Ensure every test user is in the Administrators group. OpenChoreo's
+# authz grants console access via the token's `groups` claim
+# (administrators-group-binding in wso2-ae-oc-extensions maps the group to
+# the admin ClusterAuthzRole); a user with no group has NO OC entitlement,
+# so every OC-backed project read 403s and the project view never renders
+# (#123). jq is required here, like patch-thunder-new-console.sh.
+echo ""
+GROUP_ID=$(curl -sS -H "Authorization: Bearer ${TOKEN}" \
+    "${THUNDER_URL%/}/groups" 2>/dev/null \
+    | jq -r '.groups[] | select(.name == "Administrators") | .id' | head -1)
+if [ -z "$GROUP_ID" ]; then
+    echo "❌ Administrators group not found in Thunder — OC access will 403 (#123)."
+    exit 1
+fi
+
+MEMBER_IDS=$(curl -sS -H "Authorization: Bearer ${TOKEN}" \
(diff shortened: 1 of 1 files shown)


hevayo:
no longer needed

PR #103, opened 2026-07-07
Console: Settings page — org credentials (GitHub PAT + Anthropic key) + skills catalogue (#96)

axewilledge:
## Summary
- Adds an org-level **Settings** page (sidebar-footer entry, `/settings`) with two sections: **Credentials** (GitHub PAT + Anthropic key connect/replace/disconnect against the new `/config` resource) and **Skills** (browse/create/import/sync against `/skills*`, gated on GitHub being connected since the catalogue is repo-backed).
- Frontend-first contract diff in `packages/contracts/api/v1/openapi.yaml`: types `list-skills`, `list-skill-updates`, and `sync-skills`'s request/response bodies (previously untyped `additionalProperties: {}`); also fixes `ConfigProjection.gitProvider`/`llm` to be nullable, matching the BFF's actual not-connected behavior. BE handshake: #100.
- Migrates `features/projects`'s `useGithubOrg` off the now-removed `/org/credentials/github` onto the shared `/config` query (this endpoint moved during the mid-grill contract rewrite in #97).
- No role gate in v1 (issue #96 decision) — any authenticated org member can reach Settings; there's no server-side RBAC on `/config`/`/skills*` today either. Known and explicitly accepted, not tracked separately.
- Typed MSW mocks cover not-connected/connected/validation-error for both credentials, GitHub-not-connected blocking the Skills tab, create/import/sync/delete, and load errors. Toggle scenarios via `localStorage.setItem('aep:mock:settings', 'empty' | 'connected' | 'error')`; type `invalid` into a PAT/API-key field to trigger a validation error.

## Test plan
- [x] `pnpm typecheck` (monorepo-wide) passes
- [x] `pnpm --filter @aep/console lint` passes
- [x] `pnpm --filter @aep/console build` passes; mock code confirmed excluded from the production bundle
- [ ] Manual click-through in mock mode (`VITE_API_MODE=mock`, `pnpm --filter @aep/console dev`) — not yet done by a human; screenshots to follow

Closes/continues #96. BE handshake: #100.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #100, opened 2026-07-07
aep-api: type list-skills/list-skill-updates + selective sync-skills for console Settings (#96)


axewilledge:
BE handshake for console feature #96 (Settings — org credentials + skills catalogue), per the console development flow. The spec (`packages/contracts/api/v1/openapi.yaml`) is the source of truth (#76); the FE PR will carry this diff.

`/config` (GET/PATCH), `/config/git-provider/disconnect`, and most of `/skills*` already match the feature's needs as-is — no change requested there. Two gaps in the existing `/skills*` surface block typed mocks / the chosen UX:

## 1. Type `list-skills` and `list-skill-updates` responses

Both operations currently respond `additionalProperties: {}` (untyped) even though the BE already has concrete internal types in `services/aep-api/internal/feature/skills/skill_service.go` and `reconcile.go`:

```yaml
SkillSummary:
  type: object
  required: [name, kind, version, description, contentSha, editable]
  properties:
    name: { type: string }
    kind: { type: string }
    version: { type: integer }
    description: { type: string }
    contentSha: { type: string }
    editable: { type: boolean }

SkillUpdate:
  type: object
  required: [name, repoVersion, embeddedVersion]
  properties:
    name: { type: string }
    repoVersion: { type: integer, description: "-1 when the skill has no repo row yet" }
    embeddedVersion: { type: integer }
```

`GET /skills` → an object wrapping `items: SkillSummary[]` (matching the `OrganizationList`/`ListExternalResourcesOutputBody` pattern already used elsewhere in the spec). `GET /skills/updates` → `items: SkillUpdate[]`.

Rationale: the console's mock layer (api-guidelines.md) requires typed responses — hand-rolled untyped JSON fixtures are forbidden. Both Go types already exist server-side; this is a schema-only change, no new BE logic.

## 2. Skills repo URL on the list envelope

The Import dialog's "Via pull request" guidance path links the user to the org skills repo. Only the BE knows that repo authoritatively (`ensureSkillsRepo`, `repo_store.go`). Requested: the `GET /skills` response envelope gains a `repoUrl` field (the repo's HTML URL), alongside the typed list from §1.

```yaml
# GET /skills response envelope
properties:
  skills: { type: array, items: { $ref: SkillSummary } }
  repoUrl:
    type: string
    description: HTML URL of the org skills repo backing the catalogue.
```

## Notes

[...]

Issue #96, opened 2026-07-07
Settings — org credentials (GitHub PAT + Anthropic key) and skills catalogue


axewilledge:
## Problem
The platform is non-functional without org-level configuration: a **GitHub**
credential (to read/write project spec + code repos, and to host the org's
skills repo), an **Anthropic API key** (for the platform and coding agents to
call Claude), and the org's **skills catalogue** (the instructions/skills that
shape agent behavior — a repo-backed resource living in the org's GitHub). The
BFF already exposes all of this (`/config`, `/skills*`), but the console has no
surface for it — an org can sign in (SSO, #91) and still be dead in the water,
with no way to configure the platform from the UI.

## Users
- **Architect / SRE** — owns org-level platform configuration (the PRD's Admin
  persona): sets/replaces/disconnects the org's GitHub + Anthropic credentials,
  and browses/imports/syncs the org's skills. **v1 note:** there's no
  role gate yet (see Access control decision) — any authenticated org member
  can reach Settings; Architect/SRE is the intended owner, not an enforced
  restriction.

## Experience walkthrough

Settings is an **org-level** surface, reached from the **Settings item at the
bottom of the sidebar** (`Sidebar.Footer` in `AppLayout.tsx`, already stubbed
in as "Stub until the Admin feature lands" — this feature wires it to
`/settings`). This is distinct from the separate `UserMenu.Item label
"Settings"` (a personal/profile-settings stub) — that one is untouched and out
of scope here. Settings has two primary sections: **Credentials** and
**Skills**. There is **no role gate in v1** — any authenticated org member who
can reach `/settings` gets full access (see Access control decision below).

### Credentials section
1. Renders from `GET /config` → `ConfigProjection { gitProvider, llm, idp }`
   (idp is out of scope for this feature — see Scope). Each of GitHub and
   Anthropic shows its own connected/not-connected state. No secret is ever
   returned or shown — only projections (masked prefix/last4, resolved
   identity, timestamps, status).
2. **Anthropic:** a masked API-key field. Typing a new key and submitting
   always does the same `PATCH /config` with `llm: {kind: "anthropic",
   apiKey}` — connect and replace are the same inline action, no separate
   "Replace" affordance and no confirmation dialog. The PATCH **synchronously
   validates** the key against Anthropic before persisting (probe-before-write);
   on success the section shows the connected projection (`keyPrefix`•••
   `keyLast4`, `connectedAt`, `lastValidatedAt`, `status`); on failure the
[...]

PR #126, opened 2026-07-08
fix(console): don't hang the skills Import dialog on a bodyless non-2xx (import-by-URL)

axewilledge:
Follow-up to #103 (merged) for the Settings → Skills feature (#96), from live testing.

## Problem
Submitting **Import → From URL** against the live BFF hangs the dialog silently: no error, no result, spinner just stops. Console shows `POST /skills/import-url 405 (Method Not Allowed)` — the endpoint isn't deployed yet (BE handshake #100 §3).

## Root cause
`useImportSkillUrl` threw only `if (error)`. openapi-fetch (`src/index.js`) returns `{ error: undefined, response }` for a **non-ok response with an empty body** (line 231, `Content-Length: 0`) — and a gateway 405 is exactly that. `error` is falsy, so the throw was skipped, `undefined` was returned as a success, and `onSuccess(setResult)` set the result to `undefined` → dialog stayed idle.

This is latent for any bodyless non-2xx (502/503 from a proxy, etc.), not just this 405.

## Fix
Gate on `response.ok`, not just `error`. A 405 now shows a clear inline "Import by URL isn't available on this server yet."; any other non-ok still surfaces the problem-detail message. The reject-hard → inline-error contract (grilled on #96) now holds for the unimplemented endpoint too.

Single-file change (`features/settings/api/queries.ts`). Typecheck + eslint clean. The inline-error rendering path itself was already verified in #103 (the 422 hard-reject case).

Refs #96, #103


axewilledge:
Closing this PR as we decided not to implement this feature

axewilledge:
**Superseded — closing.** This PR fixed a silent-hang bug in `useImportSkillUrl` (a bodyless non-2xx, notably the gateway's 405, left `error` undefined and slipped the `if (error)` guard).

That hook no longer exists: #143 removed the From-URL import path along with its contract surface (`POST /skills/import-url` + `ImportSkillUrlInput`), because the endpoint needed a server-side fetch the BE was never going to implement. See https://github.com/wso2/labs-agentic-engineer/issues/143#issuecomment-4923061161 and PR #145.

The underlying lesson still stands and is worth remembering: `openapi-fetch` returns `{ error: undefined }` for a non-ok response with an empty body (`src/index.js`, the `Content-Length: 0` branch), so `if (error)` alone is not a safe success check. The remaining mutations all hit endpoints that return `problem+json` on failure, so none of them are currently exposed to it.

PR #142, opened 2026-07-09
feat(console): guide credential setup on the GitHub + Anthropic cards (#96)

axewilledge:
Post-#103 UX refinement of the Settings **Credentials** cards (#96). Guidance/links only — no contract change; verified in mock mode across connected and not-connected states.

## GitHub Integration
- **Shows the org** — the connected `githubLogin` now renders as its own "Organization" line.
- **Link to create a PAT** — "Create a token on GitHub" → the fine-grained token page (present in both connect and replace flows).
- **Setup instructions emphasising org-level access** — a fine-grained token with the organization set as the resource owner, or a classic token with the `repo` scope owned by an org member.
- **Pro tip** — create a dedicated GitHub org for the platform so its agents' repos, tokens, and skills catalogue stay isolated from your team's main org.

## Anthropic Integration
- **Purpose + fallback description** (shown always): "Configure an Anthropic API key to dispatch the remote coding agent for this organization. Requirements, architecture, and task generation use the platform-provided key as a fallback if you don't configure one here."
- **Link to get a key** — "Get an API key" → the Anthropic console.

Typecheck + eslint clean on both files. Screenshots (not-connected, Anthropic, connected) to drag in below.

Refs #96

### DIFF
--- a/apps/console/src/features/settings/components/AnthropicCredentialCard.tsx
+++ b/apps/console/src/features/settings/components/AnthropicCredentialCard.tsx
@@ -73,8 +73,10 @@ export function AnthropicCredentialCard({
         <Box sx={{ display: "flex", alignItems: "center", gap: 1.5, mb: 2 }}>
           <Key size={22} />
           <Typography variant="h6">Anthropic</Typography>
-          {connected && (
+          {connected ? (
             <Chip label={llm.status} size="small" color="success" />
+          ) : (
+            <Chip label="not connected" size="small" color="warning" />
           )}
         </Box>
         <Divider sx={{ mb: 3 }} />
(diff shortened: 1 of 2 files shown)


axewilledge:
Pushed `168516b` addressing review feedback: the **GitHub org name is now required** to connect. Previously a PAT alone could connect with no org.

- Field marked required (`*`) with helper text.
- Connect / Replace stays disabled until the org is non-empty; submit always sends the trimmed org.
- When already connected, the field pre-fills from the projection so a token rotation stays PAT-only (no re-typing the org).

Client-side validation only — no contract change (`githubLogin` remains optional in the OpenAPI). Verified in mock mode: PAT-only → disabled, PAT + org → enabled, and the connected/replace pre-fill.

axewilledge:
Pushed `3be39aa`: both credential cards now show an amber **"not connected"** chip in the header when unset, mirroring the green "connected" chip — surfacing that GitHub and Anthropic are both essential for the platform to run. Verified in mock mode (not-connected shows warning chips on both cards).

PR #147, opened 2026-07-09
feat(devflow): unique base36 run suffix on dev + task workflow ids

kanushka:
## What

Give every dev-flow run a unique **base36 epoch-millis suffix** on its workflow
id, and thread the same suffix into each child task workflow id.

```
devflow-<org>-<project>-<tag>-<suffix>
taskflow-<org>-<project>-<tag>-<issue>-<suffix>
```

## Why

Follow-up to the workflow-id discussion on #135
(https://github.com/wso2/labs-agentic-engineer/pull/135#discussion_r3549065420).
That thread flagged that the dev workflow id was deterministic on
`(org, project, tag)` and relied on `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE`,
which makes re-runs of the same tag **stack under one workflow id / Temporal
timeline** and leaves a residual race on the auto-tag path.

A unique per-run suffix fixes that: re-running the same tag (e.g. after a build
failure) now produces a **distinct workflow id with its own timeline**, which is
exactly what you want when debugging a failed run.

## How

- `newRunSuffix(time.Time)` = `strconv.FormatInt(t.UnixMilli(), 36)`. Charset
  `[0-9a-z]`, compact (~8 chars), Temporal/DNS-label-safe. Millisecond resolution
  makes a collision between two re-runs of the same `(org, project, tag)`
  effectively impossible.
- The suffix is computed **once in the API handler** (`start`) from
  `time.Now()` — never inside workflow code, which must stay deterministic — and
  passed down via `DevFlowInput.RunSuffix` so a dev run and all its task children
  **share one lineage**.
- Both id builders take a `suffix` param and keep the **legacy format when it is
  empty** (backward compatible).

## Notes

- **Concurrency guard unchanged.** With unique per-run ids, Temporal's same-id
  reuse policy is no longer the dedup mechanism; the existing
  `RunningDevByProject` 409 check in `start` remains the one-dev-run-per-project
  guard, so this does not widen the race that #135 discussed.
- **Signal routing unaffected.** Webhook signals resolve the target workflow via
  DB lookup (`RunningTaskByIssue` / `RunningDevByProject`), never by parsing the
  id, so the extra id segment is transparent.

## Tests

TDD — tests written first. `id_test.go` covers suffixed + legacy formats for both
ids and the base36 encoding contract. All devflow package tests pass; `go vet`
clean.

### DIFF
--- a/services/aep-api/internal/feature/devflow/devflow_huma.go
+++ b/services/aep-api/internal/feature/devflow/devflow_huma.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strconv"
+	"time"
 
 	"github.com/danielgtaylor/huma/v2"
 	enumspb "go.temporal.io/api/enums/v1"
@@ -177,7 +179,11 @@ func (s *HumaService) start(ctx context.Context, in *startInput) (*startOutput, 
 		}
 	}
 
-	workflowID := DevWorkflowID(in.OrgHandle, in.ProjectName, tag)
+	// One run suffix per start: it distinguishes each re-run of the same tag
+	// (own workflow id + timeline) and is threaded to task children so the dev
+	// run and its tasks share one lineage.
+	suffix := newRunSuffix(time.Now())
+	workflowID := DevWorkflowID(in.OrgHandle, in.ProjectName, tag, suffix)
 	run, err := c.ExecuteWorkflow(ctx, client.StartWorkflowOptions{
 		ID:                    workflowID,
 		TaskQueue:             s.rt.TaskQueue(),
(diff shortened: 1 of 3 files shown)


kanushka:
Dropping this. Re-running a tag already works without a distinct workflow id — Temporal starts a new runId under the same workflow id (AllowDuplicate), each with its own start time and timeline, and workflow_runs is keyed on (workflowId, runId) so prior attempts are preserved. A per-run suffix only changes Temporal-list presentation, which is the same churn 5ad0e8e deliberately removed. The actual blocker earlier was the OpenBao credential path, not the id scheme.

kanushka:
Reopening — the suffix pattern `devflow-<org>-<project>-<tag>-<suffix>` is the intended design. To be clear on the relationship to #135's `5ad0e8e`: that commit dropped the human-readable `<yyyyMMddHHmmss>` datetime **prefix**; this adds an opaque base36 epoch **suffix** purely to give each re-run its own workflow id / timeline. Same as the previous state of this PR.

kanushka:
Latest changes use WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE, so re-running a tag already gets a fresh Temporal run/timeline under the same id. Closing this PR.

Issue #164, opened 2026-07-10
Build/devflow path never provisions dependencies — coding tasks with external/platform-resource deps hang forever


kaje94:
## What to build

Make `POST /projects/{p}/build` (the Temporal `DevFlowWorkflow`) provision a design's dependencies, so an app that declares any `external` or `platform-resource` dependency (a third-party API key, a database, `thunder-app` auth — the platform's core value) actually builds and deploys instead of hanging.

## The bug

The single-tag build refactor (after #135) **dropped the dependency-provisioning wiring**. A design that declares an `external`/`platform-resource` dep builds a plan whose coding Tasks **never dispatch**: the funnel gate that guards dispatch never sees those deps as satisfied, so each such Task sits `queued` and its `TaskFlowWorkflow` silently times out after 2h.

### Root cause
The provisioning wiring — (a) register `external` deps into the org catalog, (b) mint one `aep:provision` gate issue per distinct `external`/`platform-resource` dep, (c) derive end-user-auth from a `thunder-app` dep — lives ONLY in `designService.SaveAndProceed` (`internal/feature/design/design_service.go:779` `registerExternalResources`, `:784-788` `EnsureProvisionIssues`; `derive_auth.go`). `SaveAndProceed` was the old `save-design` handler and is now **de-exposed and callerless** (`design_huma.go:116-118` removed the op; `grep -rn SaveAndProceed` outside `design/` returns nothing). The new build path (`build_huma.go build()` → `artifacts.TagSpec`) runs only the structural hard gate.

### Why the tasks hang
`execution/funnel.go depsGate` (`:305`) requires every provisioning dep of a component to be "deployed", resolved via `depDeployed` (`:336`) → `view.provisionByDep[depName]`, which is populated **only from `aep:provision` gate issues**. No issue minted ⇒ `depDeployed` false forever ⇒ Task never admitted ⇒ `TaskFlowWorkflow` waits on a PR that never comes (`workflow_task.go:74`, 2h timeout). The `POST .../dependencies/{dep}/provision` endpoint still authors the real OpenChoreo Resource standalone (works even with `issueNumber==0`), so infra can be created — but nothing closes the funnel gate.

## Proposed fix
Wire the provisioning steps into the build flow so `POST /build` is self-contained. Either:
[...]

PR #194, opened 2026-07-11
fix(skills/go): build the main package, not `./...`, in the component Dockerfile

kaje94:
## What this fixes

AEP generates each user's application with **coding agents**. For a Go
service, the agent follows the `go` **skill**, which tells it how to write the
component's `Dockerfile`. That Dockerfile is later run by the platform's image
build (Argo) to produce the container that gets deployed.

The skill's canonical Dockerfile builds the service correctly:

```dockerfile
RUN CGO_ENABLED=0 go build -ldflags='-s -w' -o /out/app ./
```

but the same skill also shows a **verify** command that compiles *every*
package and throws the binaries away:

```bash
go build -o /dev/null ./...   # compile-check only
```

The agent conflated the two and emitted the build line as
`go build -o /out/app ./...`. That is invalid: `go build` cannot write
**multiple packages** to a single `-o` file. Any real service has more than
one package (`main` + `internal/handlers` + `internal/models` …), so the image
build fails and the component never deploys:

```
go: cannot write multiple packages to non-directory /out/app
Error: building at STEP "RUN ... go build -o /out/app ./...": exit status 1
→ Argo build workflow: Failed
```

## Where it breaks

```mermaid
flowchart TD
    A["Coding agent writes<br/>weather-api/Dockerfile"] --> B{"go build -o /out/app  ???"}
    B -->|"./...  (selects ALL packages)"| C["go: cannot write multiple<br/>packages to non-directory"]
    C --> D["Argo build workflow: Failed"]
    D --> E["No image → component never deploys"]
    B -->|".  (the single main package)"| F["one binary /out/app"]
    F --> G["Image pushed → component deploys ✅"]
    style C fill:#f8d7da,stroke:#c00
    style D fill:#f8d7da,stroke:#c00
    style E fill:#f8d7da,stroke:#c00
    style F fill:#d4edda,stroke:#0a0
    style G fill:#d4edda,stroke:#0a0
```

## The change

Two edits to the `go` skill (repo-root `skills/go/SKILL.md`, plus its vendored
`skills/embedded/` copy that is compiled into the aep-api binary):

1. **A positive inline comment at the build line** — the point of action, so
   the distinction is unmissable exactly where the agent writes it:

   > Build the main package: `./` (the module root, where `main.go` lives) or
   > `./cmd/<name>`. A real `-o` target takes exactly ONE package — with
   > `main` + `internal/*` present, `go build -o /out/app ./...` fails "cannot
   > write multiple packages to non-directory". (`./...` is for the
   > `-o /dev/null` verify only.)

2. **A troubleshooting-table row** mapping the exact error string → the fix,
   for the separate "I'm debugging a failed build" access path.

No new prohibition was added: stating the rule positively at the point of
action avoids duplicating it or phrasing it as a "don't" (which reads as a
negation and tends to backfire).

## How it was found

Surfaced while driving an end-to-end validation build: the `weather-api`
component's Argo build failed on exactly this line. Patching the Dockerfile to
`-o /out/app .` made the build succeed, confirming the root cause.

## Verification

- `make vendor-skills` re-run; `skills/embedded/go/SKILL.md` byte-matches the
  source (so `vendor-skills-check` stays green).
- Confirmed a real build with `-o /out/app .` succeeds on the multi-package
  `weather-api` service that previously failed with `./...`.

https://claude.ai/code/session_01Y49xNCMjuch9q9GYUfDTte

### DIFF
--- a/services/aep-api/skills/embedded/go/SKILL.md
+++ b/services/aep-api/skills/embedded/go/SKILL.md
@@ -200,6 +200,10 @@ WORKDIR /src
 COPY go.mod ./
 RUN go mod download
 COPY . .
+# Build the main package: `./` (the module root, where main.go lives) or
+# `./cmd/<name>`. A real `-o` target takes exactly ONE package — with main +
+# internal/* present, `go build -o /out/app ./...` fails "cannot write multiple
+# packages to non-directory". (`./...` is for the `-o /dev/null` verify only.)
 RUN CGO_ENABLED=0 go build -ldflags='-s -w' -o /out/app ./
 
 FROM alpine:3.20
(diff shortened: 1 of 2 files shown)


kaje94:
Folded into the comprehensive #164 PR #195 (which includes this go-skill fix plus the full build-time dependency provisioning work). Closing to keep a single PR for all the changes.

PR #236, opened 2026-07-14
feat(aepctl): local source-driven dev mode (aep dev)

senithkay:
## What

Adds an opt-in **`aep dev`** command group to `aepctl` — a local inner loop for developers editing AEP service source, so a code change can be run in the k3d cluster (or via a hot-reloading dev server) with one learnable command. All commands follow `aep dev <verb> [service]`.

| Command | Purpose |
|---|---|
| `aep dev enable [--project-path DIR]` / `disable` / `set-path DIR` | Opt into dev mode + point at a local AEP checkout (persisted to `~/.aep/config.yaml`). |
| `aep dev status` | Show mode + per-service Deployment image / ready replicas. |
| `aep dev reload <svc\|all>` | **In-cluster:** `docker build` → `k3d image import` → record override in `~/.aep/dev-values.yaml` → set-image + rollout. |
| `aep dev serve <svc>` | **Local dev server (HMR).** Capability-gated. |
| `aep dev restart <svc\|all>` | Rollout-restart only (no rebuild). |
| `aep dev logs <svc> [-f]` | Tail the service pod logs. |

Services: `aep-api`, `agents`, `aep-mcp-server`, `console`, `collab` (collab guarded — errors clearly until its chart is present).

## Design notes

- **Opt-in:** off by default; `reload`/`restart`/`serve`/`logs` refuse unless `dev.enabled`.
- **Persistence:** `reload` writes image overrides to `~/.aep/dev-values.yaml`; `aep init` applies them with `-f` when dev mode is on, so local images survive a reinstall (uses `pullPolicy: Never`; recreate the cluster → re-run `reload`).
- **`serve` is capability-gated, not a global toggle.** Only edge/leaf services (console) declare a dev server; backends are reached by other services via cluster DNS and can't run locally, so `serve aep-api` returns an educational error pointing to `reload`. `serve console` defaults to mock data (`VITE_API_MODE=mock`); `--real` proxies to the in-cluster aep-api via an auto port-forward. A prepare step (`pnpm install` + build workspace deps) runs first so the dev server resolves the `@aep/*` `dist` packages.

## Faster in-cluster builds

Reorders the `console` + `agents` Dockerfiles so app source is copied **last** (a source-only edit skips reinstall and the workspace-dep builds), and adds BuildKit cache mounts (pnpm store; Go module + build cache for aep-api). `reload` sets `DOCKER_BUILDKIT=1`. CI's buildx already honors these; the default build command is unchanged so CI/prod behavior is identical.

## Testing

- `go build ./...`, `go vet ./...`, `make license-check` — clean.
- CLI: `aep dev --help`; `enable`/`status`/`reload`/`serve` gating verified (dev-disabled refuses; backend `serve` → educational error; unknown service → valid list).
- End-to-end on a live k3d cluster: `reload console`/`aep-api` swap the running image; caching confirmed (source edit reruns only `gen`+`build`, no reinstall/redownload); `serve console` HMR.

### DIFF
--- a/apps/console/Dockerfile
+++ b/apps/console/Dockerfile
@@ -26,13 +26,18 @@ FROM node:22-alpine AS builder
 RUN corepack enable
 WORKDIR /repo
 
-# Workspace manifests first for install-layer caching (same shape as
-# services/agents/Dockerfile).
+# Workspace manifests + package manifests first, so the install layer (and the
+# workspace-dep builds below) stay cached when only app source changes. The app
+# source itself is COPY'd LAST — see the note before `COPY apps/console`.
+# The pnpm content-addressable store is a BuildKit cache mount, so a cache miss
+# on install relinks from the store instead of re-downloading (needs BuildKit;
+# `aep dev reload` and CI's buildx both set it).
 COPY pnpm-workspace.yaml package.json pnpm-lock.yaml tsconfig.base.json ./
 COPY packages ./packages
-COPY apps/console ./apps/console
+COPY apps/console/package.json ./apps/console/
 
-RUN pnpm install --frozen-lockfile --filter @aep/console...
+RUN --mount=type=cache,id=pnpm-store,target=/pnpm/store \
+    pnpm install --frozen-lockfile --store-dir=/pnpm/store --filter @aep/console...
 
 # Workspace deps whose exports resolve to ./dist must be compiled first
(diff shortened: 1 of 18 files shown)


kaje94 (changes requested):
IMO, aepctl is an end user focused tool & we are trying to use that for local development, which doesn't feel right.

senithkay:
> IMO, aepctl is an end user focused tool & we are trying to use that for local development, which doesn't feel right.

agreed

PR #241, opened 2026-07-16
feat(deployments): intelligent run-after-pull update script

KCSAbeywickrama:
## What

Adds `deployments/scripts/update.sh` — a run-after-pull driver that diffs the
checkout against the last successful run and executes **only** the setup steps
whose inputs changed, then hands off to the existing `start.sh`. It changes no
other scripts; it only invokes them.

Run from the repo root:

```bash
bash deployments/scripts/update.sh              # detect → plan → confirm → run
bash deployments/scripts/update.sh --dry-run    # preview the plan only (no side effects)
bash deployments/scripts/update.sh --yes        # run without the confirm prompt
bash deployments/scripts/update.sh --full       # run every step regardless of the diff
```

## Why

Bringing AEP back up after a `git pull` meant manually deciding which of
`make install`, `make gen`, `make build-validation-runner FORCE=1`,
`setup-aep.sh`, and `start.sh` to run — a decision fully derivable from which
files the pull changed. This automates it.

## How it works

Change set = committed range `marker..HEAD` + uncommitted tracked changes +
untracked files, so both a pull and local edits count. Each step fires only when
a changed path matches its inputs:

| Step | Fires on |
|---|---|
| `make install` | dependency manifests (`package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, workspace `*/package.json`, `go.work`/`go.work.sum`, member `go.mod`/`go.sum`) |
| `make gen` | `packages/contracts/api/v1/openapi.yaml`, `packages/contracts/schemas/*.schema.json`, `skills/**` |
| `make build-validation-runner FORCE=1` | anything under `runners/remote-worker/` **except** `plugin/**` (live hostPath overlay — no rebuild) |
| `deployments/scripts/setup-aep.sh` | its own scripts + `deployments/manifests/**` + `deployments/single-cluster/resource-types/**` (gated tightly since it SIGHUP-restarts k3s) |
| `deployments/scripts/start.sh` | always (its `docker compose up --build` rebuilds every service image from source) |

Behaviour: prints a plan with the triggering file and asks to confirm (`--yes`
bypasses); auto-starts the k3d cluster if it exists but is stopped; refuses to
run a heavy full bring-up when the cluster is missing (points at `setup.sh`);
records the last-run commit in `deployments/.aep-last-run` (git-ignored) on
success only. First run — before the marker exists — plans a full run.

## Docs

- `docs/developer-guide/run-after-pull.md` — new "One command" section; the
  manual steps remain as the reference and single-step fallback.
- `deployments/README.md` — `update.sh` added to the files table.

## Verification

All via `--dry-run` (no bring-up triggered):

- `HEAD~5..HEAD` range → all four steps + `start.sh`, correct triggering files, k3s-restart warning shown.
- Range with only doc/script changes → **only** `start.sh` (no over-triggering).
- No marker → full run.
- plugin-only change → validation-runner skipped; `src/` change → it fires.
- unresolvable `--from` → errors, exit 1.
- confirm `n` / EOF → aborts cleanly, no marker written.
- header passes `addlicense -check`.

### DIFF
--- a/deployments/.gitignore
+++ b/deployments/.gitignore
@@ -6,3 +6,4 @@ github-app-private-key.pem
 data/
 .webhook-relay.pid
 .webhook-relay.log
+.aep-last-run
--- a/deployments/README.md
+++ b/deployments/README.md
@@ -65,6 +65,7 @@ Key wiring:
 | `scripts/setup-openchoreo.sh` | Control Plane + Data Plane + Workflow Plane + Thunder |
 | `scripts/setup-aep.sh` | ClusterWorkflows + ClusterComponentTypes + Environment + AuthzRoleBindings + `.env` |
 | `scripts/start.sh` | Refresh DNS, seed kubeconfig, `docker compose up` |
+| `scripts/update.sh` | Run-after-pull driver: diff vs last run → run only the changed steps → `start.sh` (see `docs/developer-guide/run-after-pull.md`) |
 | `scripts/stop.sh` | `docker compose down` (cluster stays) |
 | `manifests/docker-build-workflow.yaml` | `dockerfile-builder` ClusterWorkflow (Argo CWTs) |
 | `manifests/aep-coding-agent.yaml` | Coding-agent one-shot pod template (mirrors v2 exactly) |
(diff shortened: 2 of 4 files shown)


kaje94 (changes requested):
IMO, we are adding extra complexity to our local runs by introducing our own mechanism to maintain build caches. 

Senith is currently working on a [skaffold](https://skaffold.dev/) based dev setup solution that should ideally fix what this update script is doing, but in a more scalable manner

PR #248, opened 2026-07-17
feat(aepctl): production-only CLI — command restructure, in-cluster config, and day-2 operations

senithkay:
## Summary

Reworks `aepctl` from a combined dev+prod tool into a **production-only Kubernetes CLI** for platform engineers operating AEP on an OpenChoreo cluster.

### Command restructure

All commands are now grouped by domain:

```
aep platform install      — full platform bootstrap (was: aep init)
aep platform uninstall    — reverse platform install
aep platform configure    — push local config YAML to in-cluster ConfigMap
aep platform update       — helm upgrade with per-service image pins
aep platform rollback     — show history + helm rollback
aep platform status       — helm release + pod table + config check

aep sre install           — install obs plane + RCA agent
aep sre uninstall         — reverse sre install
aep sre status            — helm releases + deployments + ESO secrets + pod table
```

Removed dev-only commands: `login`, `openbao`, `thunder`, `connect`.

### In-cluster config (replaces `~/.aep/config.yaml`)

`aep platform install` writes all non-sensitive config to an `aep-cli-config` ConfigMap in `wso2-aep` after bootstrapping. Subsequent commands load it via `PersistentPreRunE` — no local file, no out-of-band distribution, no plaintext secrets on disk.

- Sensitive values (`thunder.admin_client_secret`) are never stored in the ConfigMap; read at runtime from the ESO-synced `aep-thunder-secrets` Secret
- Precedence: `CLI flag > AEP_* env var > ConfigMap > code default`
- `aep platform configure --config ~/.aep/config.yaml` lets operators update cluster config without reinstalling
- `config.ConfigMapKeys` is the single source of truth shared by install and configure

### Production defaults

Three viper defaults flipped from `true` to `false` since this is now a prod-only tool:
- `oc.local_org_provisioning.enabled`
- `codingagent.local_stubs.enabled`
- `webhook.local_smee.enabled`

### Missing init flags added

Three prod-relevant flags that had no CLI surface now have flags on `aep platform install`:
- `--webhook-delivery-url`
- `--cluster-gateway-proxy-url`
- `--secret-manager-api-url`

### Scoped uninstall

`aep platform uninstall` now reverses exactly what `aep platform install` created — no more touching bootstrap resources (OpenBao, ESO, the `aep` Helm release, the namespace). Optional `--purge-config` deletes the ConfigMap.

### Local image pinning

`aep platform update` supports per-service image overrides (`--aep-api-image`, `--agents-image`, `--collab-image`, `--mcp-server-image`, `--console-image`) with `--pull-policy Never` for local images loaded via `k3d image import` / `kind load docker-image`.

### `--kubeconfig` flag

All commands now respect `--kubeconfig` (was hardcoded to `""` in several places).

## Test plan

- [ ] `go build ./tools/aepctl/... && go vet ./tools/aepctl/...` passes clean
- [ ] `aep platform --help` and `aep sre --help` show correct subcommand trees
- [ ] `aep platform install` completes on a fresh OC cluster and writes `aep-cli-config` ConfigMap
- [ ] `aep platform configure --config ~/.aep/config.yaml` upserts the ConfigMap without reinstalling
- [ ] `aep platform status` and `aep sre status` print correct tables against a live cluster
- [ ] `aep platform rollback` shows history and rolls back after a bad `aep platform update`
- [ ] `aep platform uninstall` removes only the platform chart + PVCs; bootstrap release stays
- [ ] `aep sre uninstall` removes obs namespace + cluster-scoped CRs; platform stays


kaje94 (approved):

kaje94:
Does platform uninstall deletes all PVCs in the namespace instead of only the ones related to AEP?

senithkay:
> Does platform uninstall deletes all PVCs in the namespace instead of only the ones related to AEP?

it deletes all the PVCs in the namespace that it created. Currently only the AEP creates PVCs in that namespace won't affect anything else. but will update the PR to filter only the helm managed PVCs to be deleted

kaje94:
Does AEP_* env-var override work with nested config keys(eg: thunder.url) ?

> For context, config.Init() sets viper.SetEnvPrefix("AEP") + viper.AutomaticEnv() but no env key replacer. Without viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_")), AutomaticEnv looks up thunder.url as AEP_THUNDER.URL (with a dot), which shells can't set, so only the top-level server (AEP_SERVER) resolves. AEP_THUNDER_ADMIN_CLIENT_SECRET works only because init.go reads it via os.Getenv directly. If nested env overrides are intended (the docs say they are), adding the replacer or per-key viper.BindEnv would fix it.

kaje94:
If user re-runs `platform install` command, should we warn the users(specially when users have overwritten or imported configs ) ?

senithkay:
> If user re-runs `platform install` command, should we warn the users(specially when users have overwritten or imported configs ) ?

Users will override there existing configs silently. with 80cf886d4a810d9fc93cbea5f32c9609004f9efa I have added a warning so users know what would happen and what they can do to avoid that

senithkay:
> Does AEP_* env-var override work with nested config keys(eg: thunder.url) ?
> 
> > For context, config.Init() sets viper.SetEnvPrefix("AEP") + viper.AutomaticEnv() but no env key replacer. Without viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_")), AutomaticEnv looks up thunder.url as AEP_THUNDER.URL (with a dot), which shells can't set, so only the top-level server (AEP_SERVER) resolves. AEP_THUNDER_ADMIN_CLIENT_SECRET works only because init.go reads it via os.Getenv directly. If nested env overrides are intended (the docs say they are), adding the replacer or per-key viper.BindEnv would fix it.

Concern was valid. Fixed in 50aa519df8435deff9b5d94a6ec025a0cf0809ab

PR #257, opened 2026-07-20
feat(local-dev): replace Docker Compose with Skaffold + k3d in-cluster dev

senithkay:
## Summary

- Removes the Docker Compose-based local dev stack (docker-compose.yml, start.sh, stop.sh, local-secret-manager-api) entirely
- Replaces it with Skaffold + the existing k3d/OpenChoreo cluster for a fully in-cluster dev setup
- All 6 AEP service images (aep-api, agents, collab, aep-mcp-server, console, cluster-gateway-proxy) are built locally and loaded into k3d — no GHCR push needed for local dev

## Dev flow

```bash
# One-time per cluster (after OC k3d install from openchoreo.dev)
make setup-local       # creates K8s Secrets + registers Thunder OAuth clients

# Inner dev loop
make dev-cluster       # skaffold dev: build → k3d import → helm deploy → watch
```

Edit any source file → Skaffold rebuilds only the affected service and redeploys (~45-60s).

## Key changes

| File | What |
|---|---|
| `skaffold.yaml` | Builds all 6 images locally, deploys platform chart with local values |
| `values.local.yaml` | Local overrides: localSecrets, localStubs, RWO PVCs, IfNotPresent pull |
| `values.yaml` | Added `localSecrets.enabled` flag to bypass ESO/OpenBao |
| `setup-local-secrets.sh` | One-time K8s Secret creation + Thunder client registration; idempotent (preserves stable secrets across re-runs) |
| `aep-api/deployment.yaml` | Init container waits for postgres; sslmode=disable on DATABASE_URL |
| `temporal/deployment.yaml` | Init container waits for postgres |
| `cluster-gateway-proxy/deployment.yaml` | imagePullPolicy Always → IfNotPresent |
| `Makefile` | setup-local, dev-cluster, deploy-local targets |

## Notes

- `aepctl` (`aep platform install`) is prod-only — not used in this local dev path
- The platform Helm chart is identical between prod and local dev; `values.local.yaml` only flips flags to swap out ESO/OpenBao for plain K8s Secrets
- `cluster-gateway-proxy` image (`ghcr.io/wso2/aep/cluster-gateway-proxy`) needs a CI publish job before it can be used in a prod install with `localStubs.enabled=true`

## Test plan

- [ ] Fresh OC k3d cluster: `make setup-local` completes without errors
- [ ] `make dev-cluster` → all 8 deployments reach Ready
- [ ] Console accessible at `http://console.openchoreo.localhost:8080`
- [ ] Edit a source file → Skaffold detects change, rebuilds, redeploys
- [ ] `make setup-local` re-run preserves postgres password (no auth failure on next `make dev-cluster`)


skaffold.yaml
```diff
@@ -0,0 +1,112 @@
...
+      custom:
+        buildCommand: |
+          docker build \
+            --build-context skills=./skills \
+            -f services/aep-api/Dockerfile \
+            -t "$IMAGE" \
+            ./services/aep-api
+          if [ "${PUSH_IMAGE}" = "false" ]; then
+            k3d image import "$IMAGE" --cluster openchoreo
+          fi
```
kaje94, line 51: Can you check whether its possible to pass the build context as a cli flag? i feel like the build command specially with the if condition feels messy

```
build:
  local:
    push: false
    useDockerCLI: true      # REQUIRED for cliFlags to apply
    useBuildkit: true       # REQUIRED for --build-context (a BuildKit feature)
  artifacts:
    - image: ghcr.io/wso2/aep/aep-api
      context: services/aep-api
      docker:
        dockerfile: Dockerfile
        cliFlags:
          - "--build-context=skills=skills"   # path relative to skaffold's workdir (repo root), matching prod CI's buildContexts: skills=skills
 ```

File: skaffold.yaml [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+
+    - image: ghcr.io/wso2/aep/aep-mcp-server
+      context: .
+      docker:
+        dockerfile: services/aep-mcp-server/Dockerfile
+
+    - image: ghcr.io/wso2/aep/console
+      context: .
+      docker:
+        dockerfile: apps/console/Dockerfile
```
kaje94, line 77: Right now every edit triggers a full image rebuild (~45-60s). I think we can get a much faster inner loop, and it looks different per stack. These are follow-ups, not blockers.

| Artifact | Stack | Suggested inner loop |
|---|---|---|
| agents, collab | TS (runs via `tsx` in prod, and both already have `dev: tsx watch`) | override the command to `pnpm dev` + `skaffold sync` `src/**` → ~1s reload, no image rebuild, near-zero dev/prod divergence. Keep in-cluster. |
| console | React/Vite (prod = nginx static) | prefer running `vite` on the host against the in-cluster API for native HMR (register the localhost origin in Thunder for local); keep the in-cluster image only to validate the real nginx build. Containerizing the dev server adds friction for little gain. |

Sync `src/**` only, never `node_modules`/`dist` (`.dockerignore` already excludes them). Editing a shared `packages/*` dep needs that package's own `tsc --watch`, since they're consumed as built `dist`.

deployments/helm-charts/platform/values.local.yaml
```diff
@@ -0,0 +1,62 @@
...
+# Provision per-org OC ComponentTypes locally (cloud does this via platform-api).
+localOrgProvisioning:
+  enabled: true
+  orgNamespace: default
+
+# GitHub webhooks via smee.io relay. Set webhook.deliveryURL to your smee.io
+# channel URL and set webhook.localSmee.enabled: true to activate.
+webhook:
+  localSmee:
+    enabled: false
```
kaje94, line 41: do we not need this for the local flow ?

Makefile
```diff
@@ -107,6 +107,23 @@ tools:
 build-validation-runner:
 	FORCE=$(FORCE) bash deployments/scripts/build-validation-runner.sh
 
+# ── Local in-cluster dev (Skaffold + k3d) ────────────────────────────────────
+# Run once per cluster after setup-k3d.sh. Creates K8s Secrets and registers
+# AEP OAuth clients in Thunder. Idempotent.
+#   Requires: ANTHROPIC_API_KEY env var
+setup-local:
+	bash deployments/scripts/setup-local-secrets.sh
```
kaje94: Can you check the possibility of making local setup fully declarative so that `make dev-cluster` with just `ANTHROPIC_API_KEY` set is enough, and the imperative `make setup-local` step goes away entirely? Most of what it does (Thunder clients, roles, CORS) already has a declarative form in `values-thunder.yaml`, and the rest are local secrets we can hardcode, so it feels like the whole thing could collapse into the deploy.

kaje94:
right now in setup.sh, the observability plane is enabled only based on an env variable `ENABLE_OBSERVABILITY`, can you check if there's a better way?

kaje94:
The resource-types in /deployments/single-cluster/resource-types are needed for the local testing flow. are these available in the new local flow?

kaje94:
Moving forwards, do we still need the setup.sh, setup-k3d.sh, setup-openchoreo.sh, etc and other related scripts or can those be deleted after confirming the entire flow?

senithkay:
> The resource-types in /deployments/single-cluster/resource-types are needed for the local testing flow. are these available in the new local flow?

Resource related stuff has not yet added. to the helm charts. they were updated after the initial PR was merged. will add them in a separate PR

senithkay:
> Moving forwards, do we still need the setup.sh, setup-k3d.sh, setup-openchoreo.sh, etc and other related scripts or can those be deleted after confirming the entire flow?

Some of them can be removed. but these scrips does not setup oc cluster. so users have to first create an oc and only then they can use these scripts on top of it

skaffold.yaml
senithkay: Valid. Fixed with b4e4b354

deployments/helm-charts/platform/values.local.yaml
senithkay: Users should set this via `values.local.dev.yaml` (gitignored per-dev override) after getting a smee channel — enabling it without a `deliveryURL` would cause the smee-client pod to crash-loop.

Makefile
senithkay: Will track this as a follow up

PR #329, opened 2026-07-28
feat(aep-api): make process composition importable via app.Run(Options)

kaje94:
## Why

Today `aep-api` boots only from `cmd/aep-api/main.go`. Cloud (wso2cloud) needs a **thin private Go wrapper** that reuses the same server, but swaps how OpenChoreo calls are authenticated (pass through the console user’s JWT, or use BFF M2M + org impersonation).

That is only possible if composition is an **importable public API**, not locked inside `internal/` and a hard-coded `main`.

This PR lifts that surface. It does **not** add the cloud wrapper itself (that lives in the private overlay repo).

## What changes

1. **Public `app.Run(Options)`** — `github.com/wso2/aep/aep-api/app` owns process lifecycle (load config → assemble → serve → shutdown). Callers pass behaviour differences through `Options` nil-able fields.
2. **Public `ocauth` package** — request-auth strategy interface, auth mode, and context helpers overlays need without importing `internal/`.
3. **Injectable OpenChoreo request-auth strategy** — transport asks the strategy which credential class to use per request. `nil` strategy = **all-M2M, never forward a user JWT**.
4. **OSS default stays simple** — `cmd/aep-api` wires **direct OpenChoreo + all-M2M** (`DirectOCStrategy`, M2M when `SERVICE_AUTH_*` is set, **no** `X-Impersonate-Org`). Impersonation code moves out of the public tree so an overlay can inject its own resolver.
5. **Docs** — `services/aep-api/design/composition-seam.md` + README updates for the shipped seam.

## How the seam fits together

```mermaid
flowchart LR
  subgraph public["Public repo (this PR)"]
    MainOSS["cmd/aep-api<br/>NewOSSOptions()"]
    AppRun["app.Run(Options)"]
    OC["OpenChoreo transport"]
    Strat["RequestAuthStrategy"]
    MainOSS --> AppRun --> OC
    AppRun -. injects .-> Strat
    OC --> Strat
  end

  subgraph private["Private overlay (separate PR)"]
    MainCloud["asdlc-service/main.go"]
    PAS["PAS strategy + impersonation"]
    MainCloud --> AppRun
    MainCloud -. injects .-> PAS
  end
```

### Per-request auth decision (OSS vs cloud)

```mermaid
flowchart TD
  Req[Outbound OpenChoreo HTTP request] --> S{RequestAuthStrategy}
  S -->|OSS: DirectOC / nil| M2M[Always M2M service token<br/>never pass through user JWT<br/>no X-Impersonate-Org]
  S -->|Cloud PAS overlay| Dec{User JWT present<br/>and not service-identity?}
  Dec -->|yes| User[Forward inbound user JWT]
  Dec -->|no| M2MImp[M2M + X-Impersonate-Org]
```

**Important:** there is **no runtime fallback** between credential classes (user JWT 401 does not silently retry as M2M). Same-class M2M 401 invalidate+retry is preserved.

## Options surface (high level)

| Field | Role |
|---|---|
| `AuthProvider` | How to mint/cache the BFF M2M token |
| `RequestAuthStrategy` | Per-request credential class (user JWT vs M2M) |
| `ImpersonateOrgResolver` | Value for `X-Impersonate-Org` when M2M needs it (nil in OSS) |
| `SecretsProvider` | Slot for a later secrets implementation (still today’s SM-API in OSS) |

## Out of scope here

- Private overlay module / PAS implementation
- Dockerfile / `.cicd` / cloud deploy wiring
- Secrets provider rewrite

## Test plan

- [x] `make build` / `test` / `lint` / `typecheck` / `license-check`
- [x] `make -C services/aep-api deadcode-check` (overlay-facing `ocauth` helpers marked `//deadcode:keep` with rationale)
- [x] `cd services/aep-api && go test ./...`
- [x] Local k3d E2E: OSS all-M2M project create; logs show **no** user-JWT pass-through and **no** `X-Impersonate-Org`

## Follow-up

Cloud PAS overlay (private repo) imports this package via `replace` today and will pin a merged SHA later when cloud image builds are wired.


kaje94:
Follow-up commit: public `ocauth` package + `app.Run(Options)` loads config internally so the private overlay module can import the seam without touching `internal/`. Overlay compile probe (PAS strategy stub + `app.Options`) succeeds against this tip.

PR #358, opened 2026-08-01
Make aep-api and console safe at multiple replicas

kaje94:
## Why this exists

Today several paths in `aep-api` and the console assume a **single** API pod. When two pods run behind the same load balancer, users can see:

1. A design/agent turn show **“Turn failed”** even though it is still running (stream attach hit the other pod’s in-memory buffer → 404).
2. Pods **crash-loop on boot** when both apply schema migrations at once, and a short window where two provision runs can be admitted for the same work.
3. **Duplicate** gate-issue comments / budget bumps when two pods race to finish the same execution row.
4. Dead wiring left after PR #332: a `Reevaluator` port that is never set, so it never runs.

This PR fixes those four problems in code. It does **not** change replica counts in cloud; that stays a later deploy step.

### 1. Console: retry turn-stream attach on 404

The turn narration buffer lives **in memory on one pod**. A browser that lands on another pod gets HTTP 404 before any SSE bytes. Previously that became a hard “Turn failed”. Now the client treats 404 as “try again with backoff”, and if the turn already finished it settles via the status GET.

```mermaid
sequenceDiagram
  participant UI as Console
  participant A as aep-api pod A<br/>(has buffer)
  participant B as aep-api pod B<br/>(no buffer)

  UI->>B: GET .../turns/{id}/stream
  B-->>UI: 404 (no local buffer)
  Note over UI: backoff, then retry
  UI->>B: GET .../turns/{id} (status)
  alt turn already completed/failed
    B-->>UI: terminal status
    Note over UI: settle bubble, stop retrying
  else still running
    UI->>A: GET .../turns/{id}/stream (retry may hit A)
    A-->>UI: 200 SSE (replay from index 0)
  end
```

### 2. Boot migrations: one lock, safer index swap

Two pods booting together used to run DDL concurrently. Migrations now take a **session-scoped Postgres advisory lock** so only one `database.Run` proceeds at a time. Separately, the executions admission index is created **before** the legacy index is dropped (same pattern as milestone runs), so there is never a window with no unique constraint.

```mermaid
flowchart LR
  subgraph before [Before]
    D1[DROP legacy index] --> Gap[No admission mutex] --> C1[CREATE new index]
  end
  subgraph after [After]
    C2[CREATE new index] --> D2[DROP legacy index]
  end
```

```mermaid
sequenceDiagram
  participant P1 as Pod 1 boot
  participant P2 as Pod 2 boot
  participant PG as Postgres

  P1->>PG: pg_advisory_lock(migration)
  P2->>PG: pg_advisory_lock(migration)
  Note over P2: blocks until P1 unlocks
  P1->>PG: RunAll migrations
  P1->>PG: pg_advisory_unlock
  P2->>PG: acquires lock, RunAll (idempotent)
```

### 3. Finish race: only the winner runs side effects

`Execution.Finish` is already conditional (`UPDATE … WHERE status IN (queued, running)`). The loser gets `(nil, nil)`. Call sites were ignoring that and still posting “✅ Provisioned” / “⚠️ failed” comments or notifying the build event plane. They now return immediately on a nil execution.

```mermaid
sequenceDiagram
  participant W1 as Watcher on pod 1
  participant W2 as Watcher on pod 2
  participant DB as executions row
  participant GH as GitHub issue

  W1->>DB: Finish(succeeded)
  DB-->>W1: row (winner)
  W2->>DB: Finish(succeeded)
  DB-->>W2: nil, nil (loser)
  W1->>GH: CloseIssue("✅ Provisioned…")
  Note over W2: skip CloseIssue / notifyBuildTerminal
```

### 4. Remove dead `Reevaluator` port

After PR #332 deleted the execution funnel, provisioning still had a `Reevaluator` dependency that `app.go` never wired. This PR deletes the interface, field, nil-guarded call, and test fakes.

#### Assumption for #332 authors — please veto at review if wrong

The instant “re-check blocked tasks now” nudge that port existed for is **superseded by the gate-issue-close webhook + the 60s `eventcore` sweep**. If that nudge must stay, say so and we will restore a real wiring instead of deleting the port.

## Out of scope (intentionally)

- Changing cloud `replicas:` (deploy later)
- Persisting turn-stream parts in Postgres (durable follow-up behind the 404 retry)
- Log-capture `ON CONFLICT` handling (deferred with the coding-agent job work)
- Advisory lock on the eventcore sweep (accepted 2× GitHub API cost at two replicas)

## Test plan

Local only (no cloud / no browser E2E in this PR):

- [x] `make test` / `make lint` / `make typecheck`
- [x] `services/aep-api` `make deadcode-check` (`Reevaluator` fully gone)
- [x] Console unit: turn-stream 404 retry + terminal settle + non-404 still throws
- [x] Integration / unit race proofs:

```
--- PASS: TestRun_SerializesConcurrentCallersWithAdvisoryLock
--- PASS: TestExecutionsAdmissionStatements_CreateThenDrop
--- PASS: TestRunExecutions_LegacyIndexUpgrade_AdmissionHeld
--- PASS: TestCompleteProvisionRow_FinishLoserSkipsCloseIssue
--- PASS: TestFailProvisionRow_FinishLoserSkipsComment
--- PASS: TestExecWatcher_BuildSuccess_FinishLoserSkipsObserver
--- PASS: TestExecWatcher_CodingFailure_FinishLoserSkipsNotify
--- PASS: TestExecWatcher_BuildPlainFailure_FinishLoserSkipsNotifyAndObserver
--- PASS: TestExecWatcher_BuildAuthBudgetExhausted_FinishLoserSkipsNotifyAndObserver
--- PASS: TestJobWatcher_FinishFailed_LoserSkipsNotify
```

- [x] Two-axis `/code-review` against `aep-rewrite` (Standards: no hard violations; Spec: missing loser tests then filled; no cancelled items built)
- [ ] Browser E2E (`agent-browser`) — **not run**; needs a live multi-replica (or sticky-buffer miss) environment. Planned with the cloud `replicas: 2` checks later.

### DIFF
--- a/apps/console/src/features/agent-chat/api/turns.ts
+++ b/apps/console/src/features/agent-chat/api/turns.ts
@@ -144,6 +144,22 @@ export async function getTurn(
   return data;
 }
 
+/** Thrown when the turn-stream attach HTTP call fails. `status` lets callers
+ *  discriminate a pre-stream 404 (buffer not on this replica / not yet minted)
+ *  from other failures. */
+export class TurnStreamAttachError extends Error {
+  readonly status: number;
+  constructor(status: number, message = "Failed to attach to the turn stream") {
+    super(message);
+    this.name = "TurnStreamAttachError";
+    this.status = status;
+  }
+}
+
+export function isTurnStreamNotFound(err: unknown): boolean {
+  return err instanceof TurnStreamAttachError && err.status === 404;
+}
+
 /**
  * Open the turn's SSE stream as a raw byte stream (replay from `from`, then
  * live tail). The caller iterates it with @aep/agent-stream's parseSseStream.
@@ -154,7 +170,7 @@ export async function openTurnStream(
   from: number,
   signal: AbortSignal,
 ): Promise<ReadableStream<Uint8Array>> {
-  const { data, error } = await client.GET(
+  const { data, error, response } = await client.GET(
     "/projects/{projectName}/turns/{turnId}/stream",
     {
       params: { path: { projectName, turnId }, query: { from } },
@@ -162,6 +178,8 @@ export async function openTurnStream(
       signal,
     },
   );
-  if (error || !data) throw new Error("Failed to attach to the turn stream");
+  if (error || !data) {
+    throw new TurnStreamAttachError(response?.status ?? 0);
+  }
   return data as ReadableStream<Uint8Array>;
 }
--- a/apps/console/src/features/agent-chat/runTurn.test.ts
+++ b/apps/console/src/features/agent-chat/runTurn.test.ts
@@ -16,16 +16,22 @@
  * under the License.
  */
 
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { StreamPart } from "@aep/agent-stream";
 
 // --- api/turns.js: openTurnStream/getTurn are the only calls runTurn makes.
+// Pass through TurnStreamAttachError / isTurnStreamNotFound so attach tests
+// exercise the real discriminator.
 const mockOpenTurnStream = vi.fn();
 const mockGetTurn = vi.fn();
-vi.mock("./api/turns.js", () => ({
-  openTurnStream: (...args: unknown[]) => mockOpenTurnStream(...args),
-  getTurn: (...args: unknown[]) => mockGetTurn(...args),
-}));
+vi.mock("./api/turns.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("./api/turns.js")>();
+  return {
+    ...actual,
+    openTurnStream: (...args: unknown[]) => mockOpenTurnStream(...args),
+    getTurn: (...args: unknown[]) => mockGetTurn(...args),
+  };
+});
 
 // --- @aep/agent-stream: parseSseStream is mocked to yield the parts a test
 // queues, bypassing real SSE byte parsing (irrelevant to this unit).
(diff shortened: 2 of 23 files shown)


kaje94:
## Follow-up on independent review findings

### Fixed in `31f2917f`
- **I1 (blocking):** build-fail / auth-exhausted Finish **DB-error** paths now `Notify` then `return` — no `notifyBuildTerminal`. Regression tests added.
- **I2:** concurrent `TryAdmit` during a widened CREATE→DROP window asserts exactly one admit; string-matching order test removed.
- **I3:** `slog.Info("waiting for migration lock")` + `SET lock_timeout = '60s'` before `pg_advisory_lock`.
- **M1:** stale sibling comment in `genai_service.go` aligned with the console retry wording.
- **M2:** extracted `settleFromTurnStatus` for the duplicated getTurn settle blocks.
- **M3:** `sleep()` removes its abort listener on both resolve and abort.
- **M5:** string-matching seam test dropped (behavioral concurrent test covers the order).
- **M6:** comment clarified to “follows RunMilestoneRuns”.
- **M7:** budget-exhausted WARN only runs after a successful Finish win (falls out of I1).

### Consciously not changing
- **M4** (fail-fast on unknown turnId): `getTurn` collapses *all* failures to `null` (network / 5xx / 404). Fail-fast on null would also abort legitimate multi-replica 404 retries when the status GET blips. Needs a status-discriminating `getTurn` first.
- **M8:** already had the `// lost the race` comment on the `exec == nil && err == nil` guard.
- **M9:** existing tests cover the three brief scenarios; schedule pinning / exhaustion / abort-during-backoff are polish, not blocking.
- **M11:** asserting Docker ran in CI is a platform/CI change outside this PR.

### M10
Review requested from @xlight05 (PR #332 author) for the Reevaluator-retirement assumption veto.

PR #429, opened 2026-08-11
fix(console): rejoin the collab room from scratch after a post-sync drop

hevayo:
## The bug

A markdown spec file comes back **doubled** after the collab websocket drops and reconnects.

The doc outlived the connection that filled it. The collab server treats git as the durable truth: it unloads a room on last-leave and reseeds a brand-new `Y.Doc` from HEAD on the next join. `useCollabSpec` built its `Y.Doc` once per project and kept it, so the provider's own reconnect carried the already-seeded doc back into a freshly reseeded room.

Markdown files are top-level `Y.XmlFragment`s — no key to converge on — so the two seedings merge as disjoint sets of inserts and the file's content appears twice. Non-md files hid the bug: they are `Y.Map` entries, where a colliding key resolves last-writer-wins.

## The fix

A drop that happens **after** we synced now throws the doc away and rejoins from scratch — the server's copy is authoritative.

- `epoch` is an effect dependency, so bumping it tears the old room down and joins with a fresh `Y.Doc` + provider.
- `syncedRef` gates it: before the first sync the doc is still empty and can carry nothing back, so the provider's own retry is left to do its job.
- A 1s settle window (`REBUILD_DELAY_MS`) plus the timer guard keeps a server that accepts-then-drops from spinning a full resync per round-trip.
- Cleanup clears a pending bump, so unmount or an org/project switch cannot resurrect the room.
- `onAuthenticationFailed` deliberately stays terminal — a rejected bearer will be rejected again on a rebuild.

The memoized return needs no `epoch` dependency: a rebuild always moves `status` (offline → connecting → connected), so consumers holding a fragment from the discarded doc are handed the new one.

**Known and accepted:** unsynced local edits are lost on a rebuild. That is the same content the server already discarded when it unloaded the room.

## Tests

`useCollabSpec.test.tsx` mocks `HocuspocusProvider` and records every constructed instance, pinning the four behaviors:

- a post-sync drop rebuilds with a *different* doc,
- a pre-sync drop does **not** rebuild,
- three disconnect events produce one rebuild,
- unmount cancels a pending rebuild.

## Proof of execution

New suite:

```
$ pnpm exec vitest run src/features/spec/collab/useCollabSpec.test.tsx
 ✓ useCollabSpec — rejoin from scratch after a post-sync drop (4)
   ✓ drops the synced doc and joins with a fresh one
   ✓ keeps the doc when the drop happens before the first sync
   ✓ rebuilds once per drop, not once per disconnect event
   ✓ does not resurrect the room after unmount

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Whole spec feature, on this branch rebased onto current `main`:

```
$ pnpm exec vitest run src/features/spec

 Test Files  18 passed (18)
      Tests  144 passed (144)
   Duration  12.55s
```

`eslint` is clean on both changed files.

### Unrelated, flagged not fixed

`pnpm exec tsc --noEmit` on `apps/console` currently fails on `main` as well, in `features/settings`, `features/validation`, `features/builds` and `src/mocks` — `codingLlm`, `credentialKind` and a `"revalidate"` run kind are missing from the generated contract types (looks like generated output is behind the API). Nothing in `features/spec/collab` is affected, and this PR neither causes nor addresses it.

### DIFF
--- a/apps/console/src/features/spec/collab/useCollabSpec.test.tsx
+++ b/apps/console/src/features/spec/collab/useCollabSpec.test.tsx
@@ -1,0 +1,137 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// @vitest-environment jsdom
+
+// A doc must never outlive the connection that filled it. The collab server
+// unloads a room on last-leave and reseeds a NEW Y.Doc from git HEAD on the
+// next join; markdown files are top-level Y.XmlFragments with no key to
+// converge on, so a reconnect carrying this doc's already-seeded fragments
+// merges them with the fresh seed's independent items and the file comes back
+// doubled. These tests pin the rejoin-from-scratch behavior that prevents it.
+
+import { act, renderHook } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type * as Y from "yjs";
+import { useCollabSpec } from "./useCollabSpec";
+
+interface FakeConfig {
+  document: Y.Doc;
+  onSynced: () => void;
+  onStatus: (data: { status: string }) => void;
+}
+
+const instances: FakeConfig[] = [];
+
+vi.mock("@hocuspocus/provider", () => {
+  class FakeProvider {
+    document: Y.Doc;
+    awareness = { on: () => {}, off: () => {}, getStates: () => new Map() };
+    constructor(config: FakeConfig) {
+      this.document = config.document;
+      instances.push(config);
+    }
+    setAwarenessField() {}
+    on() {}
+    off() {}
+    attach() {}
+    destroy() {}
+    sendStateless() {}
+  }
+  return { HocuspocusProvider: FakeProvider };
+});
+
+vi.mock("../../../auth/token", () => ({
+  getAccessToken: vi.fn(async () => "token"),
+  renewAccessToken: vi.fn(async () => "token"),
+  subscribeAccessTokenRefresh: vi.fn(() => () => {}),
+}));
+
+const USER = { name: "Ada", email: "ada@example.com" };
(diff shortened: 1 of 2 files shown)


hevayo:
### Review follow-up — `452a6eba`

Each finding re-checked against `@hocuspocus/provider@4.3.0` source rather than taken on description. Three were still valid and are fixed; two are deliberate and are left with a reason.

**Fixed**

1. **The old provider was left running.** `HocuspocusProviderWebsocket.onClose` emits the disconnected status and then arms its own reconnect on `configuration.delay` — 1000 ms, the same delay the rebuild used. Both timers were due in the same tick, and the rebuild won only by being registered first and by React committing the teardown before a socket could open and sync. Against a local server that race is winnable, and losing it means the old socket resyncs and pushes this doc's already-seeded fragments into the freshly reseeded room — the exact doubling this PR fixes. `scheduleRebuild` now calls `provider.disconnect()` up front, which clears `shouldConnect` so the close cannot re-arm.

2. **A rejected bearer was not terminal.** `permissionDeniedHandler` only emits `authenticationFailed`; the server closes the socket right after, and that close ran `scheduleRebuild` — so the room rejoined a second later with the same token, to be rejected again. The handler now latches for the life of the provider *and* clears a bump the close already queued, so both orderings end offline and stay there. Status behaviour is unchanged.

3. **No ceiling on the rebuild delay.** A rebuild costs the server a room load and a git seed, so a crash-looping pod would be held at one per second by every open tab. The delay now doubles per consecutive rebuild up to 30 s, and resets once a session has stayed synced for 30 s — without that reset, a server that syncs before each drop would reset the count every round and never back off.

**Not changed, on purpose**

4. **Skipping the rebuild when other peers keep the room loaded.** The server does unload on last-leave, but the client cannot know whether it was the last peer: a teammate may leave during our own outage, and the awareness set at drop time is exactly what the outage makes unreliable. Discarding unsynced keystrokes stays the conservative call, and it is already stated in the commit and the code comment.

5. **The flush rejection on a rebuild.** Once the provider is destroyed the ack can never arrive, so failing the in-flight flush is correct rather than incidental — a caller that waited would hit the 30 s timeout for the same outcome. Only the message is teardown-flavoured; rewording it is a copy change with no behavioural content, so it is not bundled here.

**Validation**

```
$ pnpm exec vitest run src/features/spec
 Test Files  18 passed (18)
      Tests  149 passed (149)
```

The suite grew by 5: `disconnect()` before the timer is armed, both authentication orderings, the doubling delay, and its reset. Checked for teeth by running them against the previous commit's source — 4 of the 5 fail there:

```
 ❯ src/features/spec/collab/useCollabSpec.test.tsx (9 tests | 4 failed)
     × silences the old provider before the rebuild is even due
     × cancels a rebuild the preceding drop already queued
     × blocks a rebuild the following drop would have queued
     × doubles the delay when a session drops again right after syncing
```

`eslint` clean on both files; `tsc` reports zero errors under `features/spec/collab` (the console's pre-existing failures in `settings`/`validation`/`builds`/`mocks` are untouched by this PR).

xlight05 (approved):

PR #439, opened 2026-08-12
fix(console): the validation page reads the loop's state, not just the verdict; plus on-platform project dev scripts

### LINKED ISSUE #437: The Validation page reports a self-healing run as 'Validation failed', and claims the run stopped
A run **mid-self-heal** — validation failed, attempts remain, the platform has filed the failures as work and dispatched a coding cycle — reads `Validation failed` on the Validation page, while the deployments board beside it correctly reads `awaiting fix` for the same run.

The chip is a mislabel. The verdict tile is worse: it states *"The run stopped here, so the milestone stays open for the fix."* The run did not stop.

## Reproduce

1. Take a project whose acceptance criteria a deployed build genuinely fails, with a validation attempt budget above 1 (`scripts/project-revalidate.sh <project> <tag> 2`).
2. Wait for the first attempt to settle `failed`. The supervisor mints one repair issue per failed criterion and dispatches an ordinary coding cycle.
3. While that coding cycle runs, open the **Deployments** page, then the **Validation** page.

**Expected:** both name the same state — the platform is repairing what validation found.

**Actual:**

| surface | shows |
|---|---|
| deployments board | `awaiting fix` (warning) |
| validation page chip | `Validation failed` (error) |
| validation page tile | `Validation failed` + "The run stopped here, so the milestone stays open for the fix." |

## Cause

The two surfaces read different fields.

- The board reads `ProjectStatus.deploy.validation`, resolved server-side by `validationStage` (`internal/projects/status_stages.go`), which consults the run's latest cycle and returns `awaiting-fix` for a live run holding a repairable verdict.
- The page reads `RunValidation.verdict` from `list-build-runs`. That is a **column** — the six verdicts, no lifecycle. `awaiting-fix` exists only in the `DeployStage.validation` enum, so the page's chip could never say it.

The page does share the label mapper (`validationView`) with the board, specifically so the two cannot drift — the comment on `headerChip` says as much. But it feeds it a different input, so the shared mapper bought nothing here. It already held `deploy.validation`, and consulted it *only* as a fallback when the verdict was empty; a `failed` verdict is non-empty, so the resolved value was never reached.

The same reasoning `validationStageFromRun` records server-side — *"rendering the bare `failed` verdict would tell a reader the version failed validation while the platform is repairing it"* — was simply never carried to the page.

## Also wrong, same root cause

- **A repeat attempt shows the previous attempt's report, not the live log.** Attempt 2 runs with a verdict already on the row, so `settled` computes true and the page opens on attempt 1's report under a chip announcing that validation is running. The first attempt correctly shows the log; the second cannot, because the page has no lifecycle input to distinguish them.
- **Contract drift.** `RunValidation.verdict`'s description states flatly that `failed` and `unreported` fail the run. Since the repeat loop landed they fail it only once the run's validation attempts are spent — so the field can hold a fatal verdict on a live run, which is exactly the case a client needs warning about. `DeployStage.validation`'s description was updated for the loop; this one was not.

## Why the board needs no fix

`deploy.validation` alone is correct there, and safe by construction rather than by luck. `useValidationCounts` gates on `COUNTABLE = {passed, partial, failed}`, so `awaiting-fix` fetches no report and the chip is the board's entire statement about validation — there is nothing on screen for it to contradict. And `deploy.validation == "failed"` is only reachable on a *terminal* run, so the counts always join against a verdict that can no longer change.

The Validation page is the inverse: its whole body **is** the report, pinned to the validation cycle's merge commit, and it renders that report mid-repair on purpose — that is what tells the reader what is being fixed. Once evidence from the run row is on screen, the chip above it has to come from the same join, or the two can disagree by a poll interval.

## Fix

One shared join, `validationState(deploy.validation, verdict)` — the lifecycle from the status read, the verdict from the run row — used by both the chip and the tile. Both halves are needed: `awaiting-fix` covers **both** fatal verdicts, so the page cannot tell from it alone whether a report exists to fetch (`failed`) or the server deliberately omitted the path (`unreported`). A lifecycle value is honoured only over `failed` and `unreported` (the two `ValidationVerdictFailsRun` repeats), so two independent 5s polls cannot pair a stale `awaiting-fix` with the newer poll's green verdict and announce a repair of something that passed.

The failed report keeps rendering mid-repair. Only the framing above it changes.

### LINKED ISSUE #438: Dev scripts to drive a deployed project: trigger a component rebuild after a hand-edit, and separate project scripts from the local plane's own
`scripts/` had no home for the loop a platform developer actually works in: **a project the platform already created is deployed, and you want to change one thing and watch the platform react** — without rebuilding the platform, and without going through a spec build.

Two things were missing, and one was in the wrong place.

## 1. A hand-edit to a project repo does nothing

Clone the repo the platform created, change code, push to the default branch — and nothing happens. Components are created `AutoBuild=false`, and while the platform subscribes to GitHub's `push` event, no handler is registered for it: the delivery is logged and dropped. So the fastest inner loop the platform has (edit the app, see it deploy) had no trigger at all short of opening a PR through an agent.

`scripts/project-rebuild.sh <project> <component>` is that trigger — the local-dev caller for `POST /projects/{project}/components/{component}/builds`. It tails the build log, then watches the deploy `AutoDeploy=true` performs off a green build (there is no deploy call to make, and none exists).

Both arguments are required on purpose: a build costs a pod and a registry push, and a mistyped component silently rebuilds the wrong thing.

Worth stating in the script and repeated here because it bites: **the build is not pinned to a commit.** Only the merged-PR fan-out pins a SHA; this clones the head of the repo's default branch at the moment the build pod runs. Push first, then run it.

## 2. The revalidate trigger lived with the platform's own lifecycle scripts

`deployments/scripts/revalidate.sh` acts on a *project*, not on the local plane, so it sat in the wrong tree — next to `setup`/`start`/`stop`/`teardown` and the runner image build, which are how you get a platform at all.

Moved to `scripts/project-revalidate.sh`, and its attempt budget became a positional argument (`scripts/project-revalidate.sh <project> <tag> <attempts>`) so it reads like its new sibling instead of needing an `ATTEMPTS=` prefix. The budget still defaults to 1, and that default is still the safe one: one attempt is spent by the first fatal verdict, which settles the run before the loop would file repair work — so the run reports and stops, touching neither the repo nor the deployment. Raising it makes a `failed` verdict into an issue per failed criterion, a coding cycle, a build and a redeploy, which is a real change to the project and stays opt-in.

## 3. `scripts/AGENTS.md` described a tree that never happened

It said *"nothing here yet"* and predicted that the cluster/stack lifecycle scripts and the runner image build would land there. They landed in `deployments/scripts/` instead, driven by the root `Makefile`.

Rewritten to state the boundary the two scripts above establish, so the next script lands on the right side of it:

> What lives here acts on **projects the platform created** — their BFF API and their GitHub repos. Nothing that builds or runs the platform itself belongs here.

## Out of scope

Neither script is a supported interface — they are local-dev callers for endpoints that already exist, and both are thin over them per the `scripts/` convention. Nothing in the platform changes.

KCSAbeywickrama:
Resolves #437
Resolves #438

Two independent concerns, both small, both against the self-healing validation loop that landed in #420.

---

## 1. The Validation page reports a self-healing run as failed (#437)

A run mid-repair read `Validation failed` on the Validation page while the deployments board beside it correctly read `awaiting fix` for the same run — and the verdict tile went past a mislabel into a false claim: *"The run stopped here, so the milestone stays open for the fix."* The run had filed the failures as work and dispatched a coding cycle against them.

The two surfaces read different fields. The board reads `deploy.validation`, resolved server-side by `validationStage`, which is cycle-aware. The page read `RunValidation.verdict` — a **column**, with no lifecycle in it. `awaiting-fix` exists only in the `DeployStage.validation` enum, so the page's chip could never say it. The shared label mapper was already in place to stop exactly this drift; it was being fed a different input.

**Fix:** one shared join, `validationState(deploy.validation, verdict)` — lifecycle from the status read, verdict from the run row — used by both the chip and the tile.

Both halves are load-bearing. `awaiting-fix` covers **both** fatal verdicts, so the page cannot tell from it alone whether a report exists to fetch (`failed`) or the server deliberately omitted the path (`unreported`) — opposite handling from one word. And a lifecycle value is honoured only over `failed`/`unreported`, the two `ValidationVerdictFailsRun` repeats, so the two independent 5s polls cannot pair a stale `awaiting-fix` with the newer poll's green verdict and announce a repair of something that passed.

The failed report keeps rendering mid-repair — it is the evidence of what is being fixed, and the coding cycle in flight has no validation log to show in its place. Only the framing above it changed:

| | before | after |
|---|---|---|
| chip | `Validation failed` (error) | `Awaiting fix` (warning) |
| tile headline | `Validation failed` | `Awaiting fix` |
| tile sentence | "…2 of 40 criteria failed — they are marked below. **The run stopped here, so the milestone stays open for the fix.**" | "…2 of 40 criteria failed — they are marked below. **The platform has filed the failures as work on this version and will validate again once the fix ships.**" |
| body | the failed report | unchanged — the failed report |

The tile's `failed` sentence now splits into evidence plus a state-dependent consequence clause. `unreported` gets its own mid-loop wording rather than borrowing that clause, because the platform files **nothing** for it: there is no failing criterion to turn into work, so the empty working set sends the run straight back to validate again. Promising a fix would name work that does not exist.

Two related fixes fell out of the page having a lifecycle input at all:

- **A repeat attempt now shows the live log, as the first attempt does.** That arm was unreachable before: attempt 2 runs with a verdict already on the row, so `settled` was true and the page opened on attempt 1's report under a chip announcing that validation was running.
- **The empty-state guard is keyed on the RAW verdict.** `running` can arrive from the status poll an interval before the cycle record reaches the run story, and counting that as "something to show" rendered an empty body under a "Validating" chip.

**The deployments board is unchanged**, and that is deliberate — `deploy.validation` alone is correct there and safe by construction: `useValidationCounts` gates on `COUNTABLE = {passed, partial, failed}`, so `awaiting-fix` fetches no report and the chip is the board's entire statement about validation. `deploy.validation == "failed"` is only reachable on a terminal run, so its counts join against a verdict that can no longer change.

### Contract + docs

- `RunValidation.verdict` said flatly that `failed` and `unreported` fail the run. They fail it only once the run's validation attempts are spent, so the field can hold a fatal verdict on a **live** run — the description now says so and points clients at `DeployStage.validation` for the lifecycle. `DeployStage.validation`'s description was updated when the loop landed; this one was missed. Regenerated with `make gen` + `make gen-api` (the Go generator is a separate target — `make gen` alone leaves `internal/gen` stale and `gen-api-check` fails).
- ADR-0013 decision 8 said the page "re-keys to the run's **verdict**", which taken literally is what produced this. Amended with the join.

---

## 2. Dev scripts to drive a deployed project (#438)

Independent of the above; two commits that predate it on this branch.

- **`scripts/project-rebuild.sh <project> <component>`** — the local-dev trigger for `POST /projects/{project}/components/{component}/builds`. A hand-edit pushed to a project repo's default branch does nothing today: components are created `AutoBuild=false`, and while the platform subscribes to GitHub's `push` event no handler is registered for it, so the delivery is logged and dropped. This is what makes the edit real. It tails the build log, then watches the deploy `AutoDeploy=true` performs. Note that **the build is not pinned to a commit** — only the merged-PR fan-out pins a SHA; this clones the default branch head at the moment the build pod runs.
- **`deployments/scripts/revalidate.sh` → `scripts/project-revalidate.sh`** — it acts on a project, not on the local plane, so it sat next to `setup`/`start`/`stop`/`teardown` by accident. Its attempt budget became positional to match its new sibling; it still defaults to 1, which is still the safe default (one attempt settles the run before the loop would file repair work, touching neither the repo nor the deployment).
- **`scripts/AGENTS.md`** said "nothing here yet" and predicted the lifecycle scripts would land there. They landed in `deployments/scripts/`. Rewritten to state the boundary the two scripts establish: `scripts/` acts on projects the platform created; nothing that builds or runs the platform itself belongs there.

---

## Verification

Contract regen, then the gates:

```
make gen && cd services/aep-api && make gen-api
apps/console: pnpm vitest run   → 74 files, 748 tests pass
apps/console: pnpm typecheck    → clean (tsc covers *.test.tsx)
apps/console: pnpm lint         → clean
services/aep-api: go build ./...                                    → ok
services/aep-api: go test ./internal/projects/... ./internal/delivery/...  → ok
```

New tests, each written against the defect:

- `validationState` — lifecycle wins over a repeatable verdict, verdict passes through for every settled state, and the stale-poll case (`awaiting-fix` + `passed` → `passed`).
- `VerdictTile` — the three consequence states for `failed`, `unreported`'s separate mid-loop wording, that a **final** verdict's sentence is unchanged by any state it is given, and that the tile leads with the state while keeping the verdict's counts.
- `ValidationPage` — one test asserting the awaiting-fix render leads with `Awaiting fix` (chip + tile), keeps the failed report and its per-criterion failure detail, and no longer says "the milestone stays open"; one asserting a repeat attempt shows the log.

**Not done, stated plainly:** no console screenshot. `agent-browser` is not installed in this environment and I did not want to install a global package to get one. The mock layer already carries a purpose-built fixture for this state (live run, `verdict: "failed"`, `reportPath` set, coding cycle in flight, `FAILED` artifacts) — `localStorage.setItem('aep:mock:validation', 'awaiting-fix')` against `pnpm dev` renders it without a live cluster. Happy to attach before/after images if a reviewer wants them.

The two scripts are exercised by use against the local stack, not by tests; neither is a supported interface and nothing in the platform changes for them.


KCSAbeywickrama:
@coderabbitai review

xlight05 (approved):

PR #458, opened 2026-08-13
feat(delivery): the platform owns deploy — autoDeploy off, promoted in dependency order, with planning moved into the run

xlight05:
Three decisions, in one causal chain: the platform takes the deploy verb away from
OpenChoreo's `autoDeploy`, discovers that a deploy it performs must also be *ordered*, and — because
the same argument applies to planning — moves planning off a detached goroutine into the run.

ADR-0017 · ADR-0018 · ADR-0019.

## 1 — The platform owns deploy (ADR-0017)

Every Component carried `autoDeploy: true`. A build's last step posted a Workload, and OpenChoreo's
controller cut the release and bound it on its own. The platform never performed a deploy; it only
observed that one had been asked for. Four consequences, all the same root cause — **you cannot order
anything around a step you do not perform**:

1. **Validation raced the deploy.** The supervisor minted its validation issue at "deployed-green",
   but the predicate that got there was builds-green. `RunPhaseBuilding`'s own comment admitted it:
   *"a green WorkflowRun means the deployment has been asked for, not that it has landed."* The
   validation agent asserted against whatever happened to be serving — possibly the previous release.
2. **A protected API served unauthenticated on first deploy.** `jwtAuth` lives in the binding's
   `traitEnvironmentConfigs`, which cannot be written before the binding exists — and the binding was
   created, rendered and served by the controller. `trait_sync.go` recorded this as accepted cost:
   *"accept the short first-deploy exposure window."*
3. **Three deploy-time effects were orphaned.** The cross-project access grant, the SPA
   `env-config.js` emit and the trait re-emit all hung off `ExecWatcher`, which sweeps `kind=build`
   execution rows — and the run loop mints none, so none of them fired for anything it built.
4. **A failed deployment failed nothing.** A binding stuck `Ready=False` minted no issue, named no
   terminal reason, and the run settled `succeeded` over a version that never came up.

**Now:** components carry `autoDeploy: false`, and the supervisor performs the deploy as a stage of
the cycle — `EnsureRelease` at the merge commit, then ONE `ApplyReleaseBinding` carrying the release
pin, the trait env configs and the workload overrides together, then a poll on each binding's Ready
condition.

- **One object, one writer.** `projects.DeploymentService` is the only writer of a user component's
  ReleaseBinding, absorbing what three services used to patch on three different triggers, each soft
  no-opping when the binding did not exist and each relying on somebody else to retry.
- **The desired state is a pure projection.** `DesiredDeploymentFor` turns design facts into both
  objects the platform owns, so the trait shape and its per-env config cannot disagree — a trait
  attached without its config does not degrade, it fails the whole render.
- **A failed deployment is now a named, recoverable failure**: one fix issue per component (carrying
  OpenChoreo's own condition reason), an ordinary fix cycle, and `deploy-budget` when nothing
  recovers it.

## 2 — The deploy has an ORDER (ADR-0019)

Owning the verb exposed the next problem: some of what the platform writes into a binding is only
knowable once *another* component is already serving, because OpenChoreo assigns an external address
when a binding renders.

A web app reads its backend's address out of `window._env_`; `src/env.ts` throws at module load
without it. Promoted alongside its backend, the SPA is published with **no config at all** and serves
a blank page. CORS is the same dependency in reverse — a protected API's allowlist is the project's
SPA origins. Graded together, those two look circular and unsolvable. Graded apart, they are not:

```
HARD  the provider's address is stamped into the consumer's own start-up config.
      Orders the deploy. Must be a DAG.
SOFT  the fact flows the other way (CORS origins, an OIDC callback registration).
      Orders nothing. Free to cycle.
```

```
builds green
   ├─ wave 0   no unsatisfied hard edge     → promote → wait Ready
   ├─ wave 1   providers now have addresses → promote → wait Ready
   └─ converge  soft facts, EMPTY commit — promotes nothing → wait Ready
   ↓
validation
```

`spec.HardConfigEdges` is the one authority for the rule, read by both the composer that needs an
address and the planner that orders around it, so the two can never disagree. Service→service is
deliberately **not** hard: OpenChoreo resolves it through its own connection mechanism, and ordering
it would refuse two services that call each other.

Two defects went with it, both observed live:

- **A soft fact vetoed a hard one.** An unresolved SPA URL — needed only to register an OIDC callback
  — withheld the `client_id` the bundle cannot start without. That, not the ordering, is what
  actually blanked the page.
- **The consumer was published knowingly misconfigured** and repaired afterwards by
  `ConvergeWatcher`, a 10-minute drift backstop silently carrying a hard dependency.

## 3 — Planning is a run phase (ADR-0018)

Planning ran in a **bare goroutine** launched from the build click, outside Temporal — the only thing
in this platform with no durability, no retry, no transient/permanent split and no history. A build
settled `plan-failed` after two ~7s TCP connect timeouts to github.com while the VM was CPU
saturated; at normal load the same container reached GitHub 20/20. The blip was the trigger; the
defect is that a blip could be fatal at all.

The click now admits the row and starts the run; the run fills its own milestone (gates, then plan)
as the first phase of the workflow. Mutex ordering is unchanged — the version is still claimed
synchronously, before anything slow. A transient failure is retried; only an answer ("the repository
was deleted") settles the version.

It also closes a pre-existing wedge: a row left non-terminal in `planning` could not be healed,
because `LiveRunForMilestone` counts `planning` as live, so the reconcile sweep skipped it before
ever reading issue counts — and the partial indexes then blocked every later build on that project.

## Supporting fixes

- **A rendering ReleaseBinding is pending, not failed.** `Ready=False` is a binding's *initial* state.
  Reading it as a verdict declared two healthy components dead two seconds after they were pinned and
  filed a fix issue for each. Only a reason that waiting cannot fix is terminal now.
- **`EnsureRelease` is idempotent by read-back, not by status code.** openchoreo-api answers a
  generate-release for an existing name with a bare 500, not the 409 the client was written against.
  Every caller is under Temporal's retry, and a deploy activity that fails on its third component
  retries all three — so the stage could never succeed again once it had half succeeded once. It was
  observed refusing the same two releases every ~100s for twenty minutes with a version stuck
  mid-stage. Required with or without the waves.

## What this deletes

`trait_sync.go` (+2 test files) — the reconcile service that answered "what has drifted" *after*
OpenChoreo had already served the component, replaced by a projection that answers "what should be
true" *before* the binding is written. `runtime_config_watcher.go` — the `env-config.js` sweep, which
existed only because a SPA's own URL resolves after its binding is Ready. Both were half-backstops
for one object; they collapse into one `ConvergeWatcher` that re-asserts the whole binding.

## Proof of real execution

Fresh project `todo6321` — todo SPA + API + Thunder OIDC + Postgres — spec → plan → code → build →
deploy → validate, on a live single-cluster OpenChoreo:

```
10:32:13  WAVE 0   pin todo-api      release=todo6321-todo-api-b96f6d6a0316-77740d4e
             ... 18s waiting for Ready ...
10:32:31  WAVE 1   compose env-config.js  keys=[API_BASE_URL TODO_API_URL THUNDER_AUTH_*]
10:32:31           pin todo-webapp   release=todo6321-todo-webapp-b96f6d6a0316-700aedd2
             ... 18s waiting for Ready ...
10:32:49  CONVERGE pin todo-api      release=""     <- promotes nothing
10:32:50  CONVERGE pin todo-webapp   release=""
```

Zero `deferring env-config.js`, zero `deploy failed`. The SPA's config was complete on its **first**
write — there was never a moment where it served without `API_BASE_URL`.

All three wiring edges landed, each at the right time:

| Edge | Kind | Written | Result |
|---|---|---|---|
| api → webapp (`API_BASE_URL`) | hard | wave 1, first write | present, never deferred |
| webapp → api (CORS) | soft | converge | the real SPA origin, not the trait's wildcard |
| webapp → thunder-auth (callback) | soft | converge | `redirectUris` registered |

The run settled **`succeeded`** (10:10:56 → 11:24:14): 2 cycles — coding (PR #5, resolving 2 issues)
then validation (PR #7, verdict `partial` on the generated app) — with `fixCycles 0`,
`conflictCycles 0`, `buildRetriggers 0`. Live SPA: `index 200`, `env-config.js 200`, bundle executes.
**Validation ran against a genuinely serving version, which is the point of the whole change.**

The auth window from consequence (2) was verified closed separately, against a live cluster: an
unauthenticated request on a component's FIRST ever serve returns **401**, and the same request with
a valid Thunder token returns 200 — because the policy was in the binding before OpenChoreo ever
rendered it.

`EnsureRelease` proved itself independently: after the rebuild the platform re-pinned
`todooo1414-todo-api-cd8d83b199ae-7d5c2c31`, the exact release name that had been 500ing, and it
succeeded. Zero deploy failures since.

## Gates

Full `aep-api` suite **62/62 green**; `deadcode-check`, `license-check` and `golangci-lint` (every
touched package) clean.

The live evidence above was captured on the build that preceded the final two-axis review. Those
fixes changed the deploy stage's *failure* paths — an unsatisfiable order now lands as
`cycleDeployFailed` rather than failing the workflow, and `releaseExists` narrowed to 5xx — not the
happy path the run exercised; both new paths have tests.

## Notes for the reviewer

- **Not Temporal-replay-safe.** The stage gained an activity call before existing ones, so a run
  already *inside* the deploy stage when the worker restarts fails on non-determinism rather than
  resuming. `workflow.GetVersion` is the sanctioned fix and is deliberately not used: it would mean
  keeping the two-pass promote alive for old histories, which is exactly the code being removed for
  wedging every run that reached it. Deploying this drains rather than migrates.
- **The ordering claim has an exact scope.** The planner orders only edges whose provider is in the
  same deploy set, on the ground that a provider outside it is already deployed. A consumer whose
  provider exists in the design but has never deployed at all would still be published unconfigured.
  Not reachable from the run loop (the deploy set comes from a fan-out that is red if any component
  failed, and a first build builds everything), but it is an assumption, not a proof.
- **The hard-edge cycle refusal is future-proofing, not a live property.** Today's rule runs
  webapp→service only and a component has one type, so the shipped graph cannot produce a cycle. The
  refusal is asserted on the edges directly so it holds for whatever edge kind is added next.
- `EnsureComponent` became an upsert — it has to be: the trait shape is frozen into the next release,
  and an existing component still carrying `autoDeploy: true` would have OpenChoreo promoting
  releases underneath the deploy stage.
- The last commit removes an unused `requirementsTagPattern` in `spec/genai_service.go`. Unrelated;
  it had been hiding because the TS half of `make lint` fails first in a worktree without node deps.

https://claude.ai/code/session_01L8V2nMfC2MeRPFCdbXnc1w


xlight05:
Thanks — six findings, all six addressed in e0f81a0e. Two were real defects.

**1. The zero-cycle settle admitting non-planning origins — confirmed and fixed.**
This was the sharpest catch. Adoption runs carry no `Tag` (`adopt.go:115`) and
`fillMilestone` returns early without one, so they never plan — yet the guard let them
settle `SUCCEEDED` on an empty working set. With GitHub's issue index lagging the label
write that started the run, that closes a milestone over work nothing dispatched.

I went slightly further than gating on `l.in.Tag != ""`: skipping the branch alone still
fell through to `!RunValidates(origin)` and settled anyway, so the **park is restored**
for runs that did not plan. Both call sites now share one predicate,
`plansItsOwnMilestone()`, because "may an empty working set be read as delivered" and "did
this run fill the milestone" are the same question, and two spellings could drift.
Regression test: `TestZeroCycleAdoption_ParksForTheLaggingIndexInsteadOfSettling`.

It also surfaced an under-specified fixture — `executePoll` used `Origin: SpecBuild` with no
`Tag`, a state the click cannot produce.

**2. `failRun` on the request context — confirmed and fixed.** Exactly the wedge class this
PR exists to remove: a cancelled client could leave the row in `planning` (non-terminal)
holding the spec mutex, and the reconcile sweep counts `planning` as live so nothing heals
it. Now `context.WithoutCancel`.

**3. `t.Fatalf` in handler goroutines — fixed, all four sites** (not just the flagged one);
`FailNow` off the test goroutine also left the request unanswered.

**4. Nil-collaborator guards — fixed.** `s.store == nil` added to both `Deploy` and
`PlanDeploymentWaves`; `Once` now repeats the guard `Run` applies.

**5. Test-name typo and 6. ADR fence languages — fixed.**

On the stale comments in `component_client.go` (the outside-diff note): fixed the two that
contradict the new model. The `ApplyComponentSpec` error text was already updated.

One I handled differently from the suggestion: `secretComponent` is a documented, deliberate
seam recording *why* per-component secret mounting is unwired. This repo's convention is to
retain unwired infra behind an explicit marker rather than delete it, so it keeps a
`//nolint:unused` carrying that reason. `golangci-lint` is now at **0 issues across
`internal/`**; full suite 62/62.

KCSAbeywickrama (approved):

Issue #459, opened 2026-08-13
Deployments surface misreports a run mid self-heal, and disagrees with the Validation page on a settled failure


KCSAbeywickrama:
Follow-up to #437, which fixed the Validation **page**. The deployments surface has the same class of defect and three more of its own: it was written when a fatal verdict ended a run, and validation now repeats.

## What it does

With `deploy.validation = awaiting-fix` (a live run repairing a failed validation):

- **The Dev environment card carries a pill reading "Awaiting fix"** — a label with no subject, sitting under a components-ready bar, so it reads as *the deployment* awaiting a fix. It is the only one of the nine states whose label does not name itself; the other eight are "Validating", "Validated", "Validation failed", "Validation?", "Validation error", "Validation skipped".
- **The verdict banner renders `This deployment's verdict: awaiting fix.`** — a lifecycle value announced as a verdict.
- **The stage note reads "The deployed system *was* checked…"** — past tense, about a version being repaired. `awaiting-fix` is toned `warning`, which maps to the settled rail state `attention`, so a note derived from the rail state claims the check already happened.
- **The criteria counts are suppressed.** `COUNTABLE` is `{passed, partial, failed}`, so the rail's fact degrades to a bare label — losing the one fact that explains the state.

And on a **settled** failure the two surfaces disagree outright:

| | |
|---|---|
| banner | `Validation failed — 4 of 6 criteria passed on this deployment.` |
| tile (Validation page) | `2 of 6 criteria failed. The run stopped here, so the milestone stays open for the fix.` |

One outcome, two voices, and the banner leads with the count that **passed** over a failure.

Separately, the stage note says "acceptance criteria" where the artifact is `specs/validation/validation-criteria.json`.

## Why it happens

`deploy.validation` and `RunValidation.verdict` answer different questions, and each surface picked one:

- `RunValidation.verdict` is a **column** — six verdicts, no lifecycle. `awaiting-fix` exists only in the `DeployStage.validation` enum.
- `awaiting-fix` **folds `failed` and `unreported` together**, so a surface holding only it cannot tell whether a report exists to fetch.

Each surface also wrote its own copy, so there was nothing forcing them to agree.

## What "fixed" means

- One sentence family, shared by the banner and the tile, composed rather than duplicated.
- The banner reads the run's verdict for the two lifecycle states, so it can tell `failed` from `unreported`.
[...]

Issue #423, opened 2026-08-10
A version's validation verdict comes from the newest run, not the run that asked


KCSAbeywickrama:
The Validation page and the project overview's deploy chip both read the **newest run** on a version and treat its verdict as the version's answer. That assumption is false, and has been since before revalidation existed.

A milestone accumulates runs across its life, and **an incident adoption never validates** — the origin gate refuses it. `settle` then stamps `skipped` on any succeeded run that never did.

## Reproduce (no revalidation needed)

1. Build a version to green, so its spec-build run records `validationVerdict: passed`.
2. Adopt any issue into that version's milestone — the ordinary SRE/RCA handoff (`aep:codingagent`, or the console's dispatch button). An `incident-adoption` run starts and settles `succeeded` with verdict `skipped`.
3. Open the version's Validation page.

**Expected:** the version still reads `passed`, with its report.
**Actual:** the page reads *"This version was not validated"*, `settled` computes false so the report is never fetched, and the deploy chip shows `skipped`. One adopted issue erases a passed version's whole validation record.

## Cause

Two different questions were asked of one list and conflated:

- **who owned the question** → by ORIGIN. Its verdict is the version's answer, including a legitimate `skipped` when no criteria were ever authored (a spec build with no oracle opens no cycle, and "not validated" is genuinely what to tell the reader there).
- **who answered it** → by CYCLES. Those carry the report pin and the logs, and they may span several runs.

Keying both on `runs[0]` collapses the two.

## Related, same root cause

Once a version can have more than one *validating* run, the log feed keyed to `runs[0]` shows only the latest attempt's logs and hides every earlier one — `RunFeed` takes a single `runId`.

## Fix

Fixed in #420 (`82f5c5ea`): the verdict comes from the newest validating-**origin** run, the report is pinned to the last validation **cycle** across the whole version, and one feed renders per run that produced an attempt, oldest first. `status_stages.go` gets the same rule through `delivery.RunValidates`, so the chip cannot drift from the supervisor.

Both regression tests reproduce without revalidation — a spec-build run plus a later incident adoption is enough — and each was verified by reverting its fix and watching it fail.

PR #465, opened 2026-08-13
Address the #462 review: one run selection for both validation surfaces

KCSAbeywickrama:
Addresses the CodeRabbit review of #462, which merged before the findings were acted on. Each was verified against the merged code first; all three were real.

## 1. The deployments hook selected the wrong run (Major)

`useValidationEvidence` read `runs[0]`. A newer **non-validating** run — an incident adoption — therefore hid the verdict and the committed report of the run that actually answered, and the deployments surface silently lost its counts and its sentence.

This is the bug **#423** fixed on the Validation page. That fix selects by ORIGIN; this hook never did:

```
ValidationPage.tsx   runList.find(r => VALIDATING_ORIGINS.includes(r.origin))   ← #423
counts.ts            runs.data?.runs?.[0]                                       ← never fixed
```

Worth stating plainly, since it looks like a merge artifact: **syncing with main would not have fixed this.** #423 touched a different file. It is the same defect written twice, not one fix missing.

So both selections move to `features/validation/lib/runs.ts` and both callers use them — one implementation instead of two that can diverge again. `lastMergedValidationCycle` likewise looks across every attempt rather than within one run: a revalidation is a later run on the same milestone, so the newest attempt is not always inside the newest run.

## 2. A first validation attempt was announced as a verdict (Minor)

With `deploy.validation = running` and no verdict on the row — **every project's first validation** — `verdictSentence` returned `""` and the banner fell through to its verdict-naming fallback:

> This deployment's verdict: validating.

That is exactly the defect #459 was filed for and the shared copy was introduced to stop, left standing in the one state that has no verdict at all. It now has a sentence of its own:

> Nothing reported yet — the validation attempt is still running.

Saying what there is to **see** rather than what is happening, because the rail's stage note beside it already says the deployed system is being checked, and two adjacent elements saying that is a restatement.

## 3. The accessibility assertions could pass while the mark stayed exposed (Minor)

- The standard test asserted only that `aria-hidden` **existed** — `aria-hidden="false"` would have satisfied it.
- The soft test — the appearance the page title actually uses — asserted only that both strings were in the DOM. Its hidden element is an **ancestor** of the matched text, so a render with nothing hidden would have passed.

Both now assert the value, the soft one through `closest('[aria-hidden="true"]')`.

## Proof

`pnpm vitest run` **812 passed** (up from 788), `pnpm typecheck`, `pnpm lint`, `make deadcode-ts-check`, `make license-check` all clean. Go side untouched.

New coverage, each pinned to the defect that motivated it:

- `lib/runs.test.ts` — a newer non-validating run with the attempt under an older one; a revalidation newer than the spec build; an in-flight cycle with no `mergeSha` skipped in favour of the last merged one.
- `verdict.test.ts` — a first attempt has a sentence; `awaiting-fix` cannot reach that path (it requires a fatal verdict) and still says nothing.
- `DeploymentsPage.test.tsx` — a first running attempt is not called a verdict.
- `StatusChip.test.tsx` — the tightened assertions above.

## Note for the reviewer

`src/generated/aep-api.d.ts` is gitignored, so a fresh branch off `main` typechecks against whatever was last generated locally. This branch needed `pnpm gen` before `typecheck` passed — unrelated to these changes (the stale errors were all in `agent-chat`), but worth knowing if CI or a reviewer sees the same.


KCSAbeywickrama:
@coderabbitai review

KCSAbeywickrama:
Pushed `15b53350` — a fourth case found while testing, same class as the others.

**Re-running validation on an already-passed version showed "Nothing reported yet — the validation attempt is still running."** over a version that had passed.

The first-attempt sentence in this PR read *"no verdict on the run row"* as *"this version has never been judged"*. True of a first attempt and of a self-heal repeat, which repeats within ONE run so the row already carries its verdict. A **revalidation** breaks it: fresh run row, enters the loop at validation with an empty verdict, while the run that delivered the version still holds `passed`.

Three changes, since the fix is not local:

- **`answeredRun`** — the newest validating run that actually holds a verdict. `validatingRun` keeps meaning "who could answer", which is what the "not validated" empty state still tests.
- **`running` now wins over any verdict in `validationState`.** Taking the verdict from the answering run without this would have left the chip reading "Validated" while the platform re-asked. The guard it replaces existed for poll skew — one interval of a stale "Validating", against a revalidation mislabelled for its whole duration. `awaiting-fix` keeps the guard.
- **Stale forms for `passed` / `partial` / `inconclusive`**, which a revalidation is the first thing that can put under a live state.

**One deviation worth flagging:** I proposed that revalidation and self-heal read identically, and that is wrong in exactly one clause. *"The implementation has been fixed and deployed"* is a fact about a **repair** — nothing is fixed between a version passing and someone asking again, and a revalidation of a *failed* version would have claimed a fix that never happened. It is now said only when the run holding the verdict is the one running again. Every other word is shared.

822 tests (up from 812), typecheck, lint, knip and license-check clean.

KCSAbeywickrama:
@coderabbitai review new changes

KCSAbeywickrama:
> `@KCSAbeywickrama` I will review the new changes.
> 
> ⚠️ Action not completed
> Already reviewed.
> 
> > Note: CodeRabbit is an incremental review system and does not re-review already reviewed commits. This command is applicable only when automatic reviews are paused.

@coderabbitai is this [15b5335](https://github.com/wso2/labs-agentic-engineer/pull/465/commits/15b533503406aa35257cafcee900f86d263c2e4f) reviewd

xlight05 (approved):

Issue #486, opened 2026-08-16
Multi-round grilling sessions — session mode, offers, and console session chrome


hevayo:
## Problem

Every interview point spends exactly one question form (`/start`; each amend branch) and the shared `grilling` skill forbids asking again — so specs are built on shallow interviews and silent assumptions. There is no way to run a proper multi-round grilling session anywhere in the product.

## Users

Developer/User — wants the agent to actually interrogate a decision area instead of surveying it once.

## Experience walkthrough

1. A **session** opens (from an end-of-turn offer, an assumption chip, a lens — those entry surfaces are separate features) by naming its scope as an **area checklist** ("Grilling Favorites: ownership · limits · privacy").
2. Rounds are **adaptive small batches**: one form of 1–4 questions — independent questions batched, dependent chains split across rounds. The current full-panel form is the round surface (sequenced takeover — accepted try-and-see).
3. **Write-as-you-go:** the turn that asks round N+1 first applies round N's settled decisions to the PRD — the doc visibly sharpens between rounds; the form header shows areas settled vs remaining.
4. Free text in any round can **override an earlier answer** — latest statement wins, reconciled on the next write.
5. Every round carries **"Finish — use recommendations"**: remaining areas get recommended answers, each tagged `*assumed*`. Navigating away **parks** the session (like today's pending form); returning resumes. No length guard — convergence bounds the session; the agent closes with a summary card (N asked, M assumed) that doubles as the next deep-dive offer.
6. `/start`'s quick pass keeps its one-form rule and ends with the **offer**: "drafted from N answers + M assumptions — grill me on: [areas]".

## Scope

**In:** `grilling` skill session mode (replacing "asking again never is allowed" with mode-dependent rules); `/start` end-of-turn offers; console session chrome (area checklist header, finish valve, park/resume); answer serialization gaining the decision-to-question link the summary needs; **spec-agent eval updates** (current rubrics assume the one-form interview).
**Out:** entry surfaces (chips, lenses, indicator — own features); design-turn usage (own feature).

## Contract changes

None expected (turn/question plumbing exists); flag if serialization needs contract surface.

## Open questions

- Multi-user semantics: the spec room is collaborative — who answers, who converges a session? (Flagged during wayfinding; decide during build or split out.)

## Decisions (wayfinder map #476)

Inherited from [Multi-round grilling session mechanics](https://github.com/wso2/labs-agentic-engineer/issues/477); progressive-deepening charter decision.

[...]

Issue #483, opened 2026-08-15
Fresh-project spec view blank-state prototype


hevayo:
## Question

Raise fidelity on the decided blank-state direction: mock a fresh project's spec view with ghost file-nav-as-stepper, skeleton PRD preview, and live narration — something to react to before it becomes a feature issue. What does 'ghost entry' look like in the Oxygen UI design system; how much narration is too much?

_Child of wayfinder map #476._

hevayo:
> *Posted during a wayfinder prototype session.*

Prototype is up: **[Spec View First Run](https://claude.ai/code/artifact/6b0adcf6-8ae5-48ed-be33-d91de56b2636)** — an interactive mock of the fresh-project blank state in console styling, three tabbed moments:

1. **Arrival (0–25 s)** — live narration replaces the greeting + canned chips; ghost file-nav entries name every upcoming artifact (nav-as-stepper); skeleton PRD with shimmer sections.
2. **Questions arrive** — sequenced takeover with the session's area checklist in the form header; "Skip questions" renamed to "Finish — use recommendations".
3. **PRD streams in** — the prd.md ghost solidifies with a writing pulse, the architecture ghost re-labels to "next — after you review the PRD", and the agent's narration checklist mirrors the doc outline.

Reactions on this issue or in session decide the resolution.

hevayo:
> *Resolution recorded during a wayfinder prototype session.*

## Resolution — reference design accepted, with a BE-kickoff amendment

**The prototype stands as the reference design** for the fresh-project spec view: **[Spec View First Run](https://claude.ai/code/artifact/6b0adcf6-8ae5-48ed-be33-d91de56b2636)** — three states: arrival (live narration replaces greeting + canned chips; ghost file-nav entries as the journey stepper; skeleton PRD with shimmer sections), question takeover (area checklist in the form header; skip valve renamed "Finish — use recommendations"), and PRD streaming (ghosts graduate with a writing pulse; the next stage's ghost re-labels forward; narration mirrors the doc outline). Build may polish visuals; the structure is decided.

**Amendment from the prototype reaction — the blank state's root cause gets fixed too. `/start` is BE-initiated at project creation**, replacing the console's client-side injection:

[...]

Issue #463, opened 2026-08-13
Chat rehydrate renders the composed turn prompt as the user's bubble — /start shows the whole skill body


hevayo:
## Symptom

Open a project after running `/start` (or refocus the tab — the D6 rehydrate replaces the log): the first **user bubble** in the agent chat is the raw composed model prompt, starting with

> The following skill guidance is ALREADY LOADED for this turn — apply it directly and do NOT call loadSkill… ## Skill: start … # Start …

i.e. the full `start` SKILL.md (~9.5KB in a fresh test project), rendered as if the user typed it. It is a class, not a one-off: the answers turn rehydrates as `Existing files: … Instruction: Answers: …` — every composed prompt in the thread renders verbatim as a user bubble.

## Cause

Two features interact:

- The agents service persists the **composed prompt** as the conversation's user message — `run-conversation-turn.ts` builds `prompt: note + eagerSkillsBlock + buildPrompt(files, instruction)` and `runTurn` appends it to `conv.messages`, which is what `store.save(conv)` persists. Since `bb525bbd` (*eagerSkills — inline known-needed skill bodies into the turn prompt, #335*), that prompt deliberately carries full skill bodies in the **per-turn user message** (to keep the system prompt byte-stable for prompt caching).
- The console's rehydrate, `projectableHistory` (`apps/console/src/features/agent-chat/history.ts`), projects **every** user-role message's text verbatim into a user bubble — it assumes user messages are user-typed text.

`8470b57b` (*the chat thread belongs to the project, not the browser*, #430 / PR #451) is what surfaced it: the project's one thread is now server-rehydrated for anyone opening the project, so the stored raw prompts became visible where the local, typed-text log used to be.

The underlying gap: the store has no separation between *what the model was sent* and *what the user said* — the composed prompt **is** the user message, and the projection layer trusts it.

## Reproduce

1. Create a project, run `/start`, answer the form.
2. Reload the project page (or refocus): the first user bubble is the skill body; the answers bubble carries the `Existing files:` / `Instruction:` framing.
3. Server-side: `GET /projects/{p}/agents/{conversationId}/messages` — the first `user` message begins with the eager-skills preamble.

## Fix directions

- **Projection-side strip** (cheap): `projectableHistory` drops the eager-skills block and `Existing files:` framing and renders only the user-meaningful tail (or skips composed messages entirely). Parses framing by convention.
[...]

PR #493, opened 2026-08-16
feat(console,aep-api): spec view first run — BE-initiated /start, ghost nav, skeleton PRD (#485)

### LINKED ISSUE #485: Spec view first run — BE-initiated /start with live narration, ghost nav, and skeleton PRD
## Problem

A fresh project's spec view is dead air: the console injects `/start` client-side after the view mounts (localStorage prompt), so the user stares at a canned greeting ("Hi! I'm your Agent") with off-domain suggestion chips and a disabled input for ~25s until questions appear. Nothing explains what will happen or where they are in the journey. The injection itself is fragile (the double-`/start` bug class, #432).

## Users

Developer/User — lands on a brand-new project's spec view for the first time.

## Experience walkthrough

1. **`POST /projects` with a prompt starts the `/start` turn server-side at creation** — the turn is already running while the user is on the overview.
2. On the **overview**, the Spec pipeline card shows the live state ("interviewing — 4 questions waiting" replaces "Generate spec"); opening the **agent chat rail** anywhere shows the turn streaming (narration; then the existing "N questions — Answer them →" banner deep-linking to the spec view). Questions are only ever answered on the spec panel.
3. Opening the **spec view**: live narration in the agent panel from second one ("Reading your idea… ✓ Consulted org defaults… Planning the interview"); the **file nav pre-renders the journey as ghost entries** (prd.md "written after your answers"; architecture/design/security/components "derived from the PRD"; validation criteria "minted after design"); the main pane shows the **skeleton PRD** (ghost section outline with shimmer).
4. As the PRD streams, ghosts graduate: prd.md turns solid with a writing pulse; the architecture ghost re-labels "next — after you review the PRD". Narration mirrors the doc outline section by section.
5. Canned empty-state chips are removed on fresh projects; any chips derive from the create prompt.

Reference design (three states, accepted): [Spec View First Run](https://claude.ai/code/artifact/6b0adcf6-8ae5-48ed-be33-d91de56b2636).

## Scope

**In:** BE auto-start at create; attach-anywhere turn visibility (overview card state, chat rail); ghost nav-as-stepper; skeleton PRD; narration; chip removal.
**Out:** session mechanics (its own feature); question-form changes beyond arrival.

## Contract changes (BE handshake required)

- Project create kicks off the `/start` turn using the stored prompt — **only when a prompt exists**, and **idempotent: one `/start` per project, enforced server-side** (retires the client-injection path and its double-start guard).

## States to design

Arrival narration · ghosts (dashed) vs solid files · graduating ghost relabel · overview Spec card interviewing state · chat-rail banner vs narration.

## Decisions (wayfinder map #476)

Inherited from [Fresh-project spec view blank-state prototype](https://github.com/wso2/labs-agentic-engineer/issues/483); charter blank-state decision on the map.

hevayo:
Closes #485. Reference design: the accepted three-state prototype from #483.

## The problem

A fresh project's spec view was dead air. The console injected `/start`
client-side after the view mounted (from a localStorage prompt), so the user
stared at a canned greeting ("Hi! I'm your Agent") with off-domain suggestion
chips — "Add acceptance criteria to the spec", before a spec existed — and a
disabled input for ~25 s. Three empty file groups read "No files yet"; the main
pane read "Select a file to view its content." Nothing said what was happening
or where the user was in the journey. The injection itself was fragile: it is
the double-`/start` bug class of #432.

## What shipped

**BE-initiated `/start` (`services/aep-api`)** — `POST /projects` starts the
spec interview server-side. The projects domain fires a new `specKickoff` port
(async, best-effort) after stamping the descriptor; the spec domain's
`KickoffSpec` claims a `spec_kickoffs` row (insert-on-conflict, the #420
admission pattern) so there is **exactly one auto-start per project**, stands
down when any turn already ran (`HasAny` — the #432 skip-valve bug class, now
closed server-side), retries while the repo provisions, and starts the bare
`/start` turn on the project's current thread so the idea rides in from the
descriptor exactly as a typed command's would. **Prompt-gated**: a project
created without a prompt starts nothing and keeps the manual Generate-spec
path. Contract: `CreateProjectRequest.prompt`'s description now states the
auto-start + idempotency handshake (no schema change); `gen-api` rerun.

**Server-sourced interview state (`useSpecInterview`)** — because the turn is
usually already running when the user first lands, the state cannot come from
the local chat log (`useAgentEngaged` sees nothing on a fresh browser).
`running` comes from the active-turn poll; `questionsWaiting` prefers the live
chat log and falls back to a rehydrate of the current thread — a turn that
*ended* on its `ask_questions` call leaves no active turn, so on a fresh
landing the thread is the only place the pending questions exist. Disabled on
settled projects, which pay no polling.

**Overview Spec card** — shows "interviewing…" / "interviewing — 4 questions
waiting" in place of Generate spec, and offers "Continue spec" **without** the
`?generate` signal (injecting a second `/start` into the running interview is
the whole bug). Generate spec survives only as the prompt-less project's manual
fallback.

**Spec view first run** — the file nav is now the journey stepper: empty groups
render **ghost entries** naming each upcoming artifact with a when-label
(prd.md "written after your answers"; architecture / design.md · security.md /
components "derived from the PRD"; criteria "minted after design"). Ghosts
**graduate** as files land, and the labels point forward — once requirements
exist the architecture ghost re-labels to "next — after you review the PRD".
A graduated row **pulses** while the agent is still streaming that file; the
signal needed no new plumbing, since a live-room-only entry in SpecView's file
union carries an empty sha (git has nothing to hash yet), gated on an agent
peer so the pulse stops with the turn. The main pane holds the **skeleton PRD**
— the document's own section outline with shimmer where prose will land.

**Chip removal** — the chat panel's empty state drops the greeting and the
canned chips on a spec-less project and says what is actually happening
(narration streams here; questions are answered on the spec view). Greeting and
chips are intact for a project that has spec files.

**Error states stay honest** — a failed derivation shows "Derivation failed"
instead of ghosts, and suppresses the skeleton. A shimmer promising a document
that is not coming would be worse than the void it replaced.

## Verification

`apps/console` — full suite, on this branch:

```
 RUN  v4.1.10 …/build-485/apps/console
 Test Files  82 passed (82)
      Tests  825 passed (825)
   Duration  31.27s
```

The seven new tests, scoped:

```
$ pnpm vitest run src/features/spec/components/SpecFileList.test.tsx \
    src/features/agent-chat/useSpecInterview.test.tsx \
    src/features/agent-chat/components/AgentChatPanel.test.tsx \
    src/features/projects/components/OverviewPipeline.test.tsx
 Test Files  4 passed (4)
      Tests  47 passed (47)

$ pnpm vitest run src/features/spec/components/SpecView.test.tsx \
    src/features/spec/components/SpecFileList.test.tsx
 Test Files  2 passed (2)
      Tests  32 passed (32)
```

What they pin: `useSpecInterview` reports `running` from the active poll on a
fresh landing and never rehydrates behind a running turn; it counts waiting
questions from the server thread with no local log, defers to the live log the
moment one exists, and is fully inert when disabled. The overview card swaps
Generate spec for "interviewing…" / "N question(s) waiting", navigates without
`?generate`, and arms the hook only pre-spec. The panel drops greeting + chips
on a fresh project, narrates a running turn, and keeps both for a project with
spec files. The nav renders five ghosts on a fresh project, graduates prd.md
and re-labels the architecture ghost, renders none for a group with real files,
and says "Derivation failed" instead of ghosts on failure. SpecView shows the
skeleton when no file exists, suppresses it on a failed derivation, and pulses
a room-only file — but not a committed one, and not after the agent leaves.

`services/aep-api`:

```
$ go build ./...      # clean
$ go test ./internal/spec/... ./internal/projects/...
ok  github.com/wso2/aep/aep-api/internal/spec              12.179s
ok  github.com/wso2/aep/aep-api/internal/spec/files         0.020s
ok  github.com/wso2/aep/aep-api/internal/spec/genaiturns    0.020s
ok  github.com/wso2/aep/aep-api/internal/projects           5.034s
ok  github.com/wso2/aep/aep-api/internal/projects/activityfeed 0.026s
```

Gates: `pnpm exec tsc -b --noEmit` clean, `pnpm lint` (eslint) clean,
`make license-check` exit 0.

## Live verification (2026-08-17, local single-cluster stack)

Driven in a real browser against the deployed branch (merged with #486 so both
features ran together). This replaces the earlier "evidence is test-level" note.

- [x] **First run, end to end.** Fresh project → the overview stays put (no
      auto-navigation), the Spec card reads `Agent is looking at your idea`
      and then `Agent has 4 questions about your idea`, and the CTA
      (`Continue spec`) leads to the spec view with the chat already open.
      The chat reads as one conversation: `/start` → `Looking at your idea…`
      → `I have some clarifications about your idea before I write the PRD.`
      → the questions banner. Answering shows
      `The agent is drafting the PRD…` in the pane and `Answered 4 questions`
      in the chat; the PRD streams in and the nav ghosts graduate.
- [x] **The Generate/kickoff race is gone.** Clicking the CTA immediately
      after creation — the window that previously produced *"An agent turn is
      already running for this project"* — completes normally. Database after
      the run: `spec_kickoffs.status = started` and exactly **one**
      `agent_turns` row, where the console previously fired a second `/start`.
- [x] **`spec_kickoffs` migration on an existing deployment.** Applied cleanly
      to the pre-existing dev database (previously exercised on a fresh DB
      only).
- [x] **Agents-down failure and recovery.** With `aep-agents` stopped, a new
      project's card flips from the pending line to the real failure, quoting
      the dispatch error, with a **Retry**. Retry while the service is still
      down honestly re-reports the failure instead of claiming success. After
      restarting the service, Retry starts a new turn and the project recovers
      to `Agent has 4 questions about your idea`. Turn history for that
      project: two `failed` rows, then `running`.

## Left for review / follow-up

- [ ] **Prompt-less project path not verified live.** The kickoff now starts
      for every project and the agent is expected to open by asking what to
      build; that path has unit coverage but no live run.
- Narration mirrors the doc outline only as far as the agent's own streamed
  steps do; the prototype's explicit section-by-section checklist is agent-side
  wording, not console state, and is not attempted here.
- `KickoffFailureReason` carries the raw cause (capped at 300 chars) so the
  card can name what failed. Deliberate for a platform-engineer console, but
  it is the one place raw server error text reaches the UI — easy to swap for
  a fixed sentence if reviewers prefer.


hevayo:
Feedback round from live testing:

1. **Question cards never render in chat.** The feed's only question surface
   is now pinned by regression tests at both `TurnBlock` and `AgentChatPanel`
   level (question text/options asserted absent from the chat DOM). The
   banner ("The agent has N questions — Answer them →") navigates to
   `/projects/:project/spec` from any page the chat panel is open on.
2. **Chat narrative sequence.** A `/start` turn's question banner is now
   preceded by a console-rendered line — "I have a few questions to clarify
   before drafting the PRD." (singular for one question). `buildFeed` flags
   `/start`-initiated turns (word-boundary match, survives rehydrate since
   #463 keeps the verbatim command); `TurnBlock` inserts the line once per
   turn, ahead of the first question banner, for `/start` turns only.
3. **Spec-view question form no longer depends on the chat rail.** Root
   cause: `useRoomQuestion` mirrors the *local chat log* into the collab room,
   and only the chat panel ever populated that log. New `useChatLogBootstrap`
   (mounted by `SpecView`) seeds an empty log from the server thread on
   arrival — the form now appears immediately; an attached panel's live log
   always wins over the bootstrap (re-checked at write time).
4. **Chat auto-opens beside the doc.** Entering the spec view while the
   first-run turn is active (streaming or questions pending) now requests the
   panel open, via a new chat-open counter in `chatStore` (`SpecView` and
   `AppLayout` are sibling subtrees with no shared state). Only an *increase*
   in the counter for the *current* project opens it, so a panel the user
   closed stays closed, and ordinary visits (the interview hook disarms once
   files exist) never auto-open.
5. **Skeleton PRD replaced by a working state.** `PrdSkeleton` (the mock
   document outline) is deleted — dead-code gate clean. The main pane now
   shows a small spinner + one live stage line naming the agent's actual
   stage: "The agent is preparing your questions…" while the interview runs,
   "The agent is drafting the PRD…" once questions are asked and answered.
   Sourced from the same `useSpecInterview` state the chat narration uses
   (which gained a `drafting` phase); disappears the moment real content
   streams. Ghost file-nav entries are untouched.
6. **Question form title**: "Quick questions" → "A few clarifications to
   write your spec" (the "everyone answers together" subtitle is unchanged).
7. **Overview Spec card** now covers the drafting phase and matches the new
   wording: "Agent is preparing your questions…" / "interviewing — N
   questions waiting" / "Agent is drafting the PRD…".

**Verification this round** (`apps/console`): `pnpm vitest run` — 85 files /
858 tests passed; `pnpm exec tsc -b --noEmit`, `pnpm exec eslint src`,
`make deadcode-ts-check`, `make license-check` all clean. This was
test/type/lint verification only — no new live-stack run, so the PR body's
end-to-end checklist is left as-is (unchecked) rather than checked off
without having actually re-verified against a running stack. Note the first
checklist item's phrasing ("...spec view opens onto ghosts + skeleton") is now
stale: the skeleton PRD is gone, replaced by the working state described in
(5) above.

hevayo:
Second live-testing round: four defects reproduced in the browser, four fixed on `feat/spec-first-run-485`. Root cause per item, then the tests.

### 1. "Toggle agent chat" navigated to the spec view

**Root cause — not the auto-open work, and not new to this branch.** `AgentChatPanel`'s pending-question effect (`AgentChatPanel.tsx`, introduced with the spec-form spike, well before this PR) navigated to the spec view for **any** question the thread still had pending, deduped only by `toolCallId` in a module-level set. The chat rail is mounted with `<Collapse unmountOnExit>`, so **opening the rail mounts the panel** — and on the overview the panel is unmounted while the question streams in, so the toggle click was always the effect's first sighting of that question. The result: the toggle changed route. Last round's auto-open work didn't cause it; it made the panel open in more places, which is why it surfaced now.

**Fix.** The intent ("take the user to a question the moment it *arrives*") was right; the signal was missing. `upsertQuestionMessage` is called **only** by `runTurn`'s live fold — every other way a question card enters the log (thread rehydrate, spec-view bootstrap, persisted log) goes through `replaceMessages`. The store now records arrivals there (`isLiveQuestion`), and the panel moves the user only for such a card. A rehydrated/persisted question renders its banner and waits to be clicked.

Secondary path closed: `AppLayout` read `?generate` on **any** project route, and consuming it navigates to the spec route — a stray or stale param was a second way for mounting the panel to change the route. It is now spec-route only.

### 2. The drafting working-state never appeared

**Root cause, two parts.** (a) `useSpecInterview.running` came only from the 12 s `GET /turns/active` poll, so the turn the user had just started by submitting answers was invisible to every surface until the next tick. (b) The spec pane's working branch was gated on `interview.running`, so both beats either side of it — question landed but not yet mirrored into the room form, and answers submitted but the next turn not attached — fell through to the pre-existing "Select a file to view its content." empty state.

**Fix.** `useAgentChat` publishes its own in-flight state through the store (in-memory, cleared when the turn settles and on unmount — it can never strand a spinner), and `useSpecInterview` ORs it with the poll. The pane's branch now covers the whole first-run turn, with three stage lines: "The agent is preparing your questions…", "Opening the agent's questions…" (questions landed, form a beat behind), "The agent is drafting the PRD…".

### 3. Blank grey slab on the overview while status loads

**Root cause.** `ProjectOverview` rendered a single `<Skeleton variant="rounded" height={96} />` for the entire pipeline while `useProjectStatus` was pending, and the identity line simply omitted the status chip and repo link.

**Fix.** `OverviewPipelineSkeleton` renders the journey's real frame — three stage cards with icons and titles, chevrons between them — and a skeleton where each card's version chip and state line will land; the identity line gets matching placeholders. Structure is known without asking the server; nothing the status decides is invented.

### 4. Machinery leaked into the chat as user messages

**Root cause.** Answers and the finish valve are sent as plain-text instructions (that is the contract with the agent), and `MessageList` printed `message.content` verbatim — so a submitted interview rendered as a raw `Answers:\n- "…": …` dump, and the finish valve as three paragraphs of directives to the model.

**Fix — render-side only; what is sent is untouched.** `userMessageSummary.ts` reads the two known shapes (`@aep/agent-stream`'s `ANSWER_PREFIX` / `ANSWERS_PREFIX`, and #486's `buildFinishInstruction` via its `FINISH_PREFIX` marker) into one line — "Answered 3 questions", "Finished — answered 2 questions, applied recommendations to 3 more" — with **Show details** revealing the verbatim instruction. Anything it cannot read renders exactly as before, so a reworded builder degrades to today's behavior instead of hiding a message. The finish marker is a local literal because `buildFinishInstruction` lives on the #486 branch; there is a comment to swap in the import when the two meet.

### Tests

`apps/console` — 87 files, 887 tests pass; `pnpm typecheck` and `pnpm lint` clean; `make license-check` and `make deadcode-ts-check` clean.

New/extended coverage:
- `layouts/AppLayout.test.tsx` (new) — the regression test for #1 at the level it broke: overview route, a question pending, click "Toggle agent chat" → the rail opens and `navigate` is never called. Reverting the one-line guard fails it.
- `components/AgentChatPanel.test.tsx` — no navigation when the panel opens over an already-pending question, none across close/re-open, and it *does* navigate exactly once when a question arrives on the live stream.
- `chatStore.test.ts` — live-arrival marking (and that `replaceMessages` is not an arrival); local turn activity publish/clear/isolation.
- `useSpecInterview.test.tsx` — running/drafting from a turn this tab started, without the poll; clears when it settles.
- `SpecView.test.tsx` — the working state holds while questions wait with no turn streaming; the drafting line beats the "Select a file…" empty state.
- `OverviewPipeline.test.tsx` — the skeleton names the three stages, invents no status, and marks itself `aria-busy`.
- `userMessageSummary.test.ts` + `MessageList.test.tsx` — the summaries (fed from the real builders' output), the expand/collapse, and verbatim rendering for human messages and unrecognized shapes.

hevayo:
Second live-testing round: drove the deployed branch in a browser (fresh project → interview → PRD). Four defects found and fixed; root cause named for each.

**1. The chat toggle navigated to the spec view** (`ec615420`). Clicking "Toggle agent chat" on the overview with a question pending landed the user on `/spec`. The rail is mounted with `unmountOnExit`, so opening it *mounts* `AgentChatPanel` — and the panel's "take the user to the question" effect fired for any question the thread still had pending, not only one that had just arrived. On the overview the panel is unmounted while the question streams in, so the toggle click was always its first sighting. The missing signal was arrival: the store now records questions folded live by `runTurn`, so the effect fires for a card that arrived on the wire in this tab, not for one merely found pending on mount.

**2. The drafting working-state never appeared** (`57d0cf88`). After submitting answers the pane fell back to the pre-existing "Select a file to view its content." — a finished, empty workspace while the agent was writing the PRD. Two causes: `useSpecInterview.running` came only from the 12 s active-turn poll, so a turn the user had just started stayed invisible until the next tick (`useAgentChat` now publishes in-flight state through the store and the hook ORs it with the poll); and the pane's working branch required `interview.running`, so the drafting beat could not win over the empty state.

**3. The overview's first seconds were a grey slab** (`a10b02ee`). While the status query is in flight the pipeline area rendered one blank rectangle. It now renders the journey's real frame — three stage cards with icons and titles, skeletons where each version chip and state line will land, same for the identity line's chip and repo link. Structure is known without asking the server; nothing the status decides is invented.

**4. Machinery instructions leaked into the chat** (`e8c29210`). Submitting an interview printed a raw `Answers:\n- "…": …` dump as the user's message, and the finish valve printed three paragraphs of directives to the model. Render-side only — what is sent to the agent is untouched. The feed reads the two known shapes (agent-stream's answer builders, and #486's finish instruction by its marker) into one human line — "Answered 5 questions", "Finished — applied recommendations to 6 remaining questions" — with "Show details" revealing the verbatim text. Anything it cannot parse renders exactly as before, so a reworded builder degrades to today's behavior rather than hiding a message.

Verification is still test-level; the live-verification boxes stay unchecked until this round is re-driven against the stack.

hevayo:
Third live-testing round — first run as one conversation:

**A. Nothing navigates the user.** The chat panel's arrival-triggered jump to the spec view is gone, along with the live-question bookkeeping in `chatStore` that existed only to arm it. No effect anywhere routes to the spec view now — it is reached by the overview's Spec card CTA or the chat's questions banner, both unchanged. Regression coverage in two places: `AgentChatPanel.test.tsx` ("no effect ever navigates") and, end-to-end through the real layout + panel, `AppLayout.test.tsx` — a question streaming in while the user reads the overview with the rail open leaves the route where it was.

**B. Chat rail visibility.** Unchanged and re-pinned: the overview never opens the rail by itself (new `AppLayout` test), the spec view still requests it during the first run (`SpecView.test.tsx`, untouched and green), and opening it manually on the overview shows the same conversation.

**C. Spec card wording.** "Agent is processing the idea" while it reads, "Agent has N questions" once they land (singular for one), "Agent is drafting the PRD…" after. The spec view's own working state now says the same words for the same stage — the user clicks through from that card, and "The agent is preparing your questions…" was the same beat in a second voice. The CTA had a hole worth naming: it read the *momentary* signals, so in the gaps where neither is true — between the 12 s active-turn poll's intervals, and between one first-run turn ending and the next attaching — it flipped back to "Generate spec" and re-armed `?generate`, offering to inject a second `/start` into the open exchange. `useSpecInterview` now also reports `started` (the project's thread has been spoken in at all), which is the durable form of the same fact, so "Continue spec" holds for the whole first run.

**D. The chat reads as a conversation.** The `/start` turn opens with the agent speaking: *"Looking at your idea to generate the product requirements document…"*. It is console-rendered because the stream has no equivalent — the model goes straight to reading files. It is also said in the working tail, for the window where the turn has produced nothing yet to hang a turn block off, so the beat neither arrives late nor appears twice. The hand-off line before the questions banner takes your wording: *"I have a few more questions before generating the PRD."* A single question on a `/start` turn is the brief-gathering one (see E), so it gets a neutral variant rather than claiming there were earlier questions. `AgentChatPanel.test.tsx` asserts the whole order: `/start` → reading line → narration → hand-off line → banner.

**E. Projects created without an idea.** The BE kickoff is no longer prompt-gated — `POST /projects` fires `KickoffSpec` for every project. The prompt-less path was already reachable and is verified: `turnSpecFor` attaches no idea, `prompts/turn.ts` appends nothing, and `skills/start` opens with its one `ask_question` ("What are you building?"). The console renders that first question through the same room-mirrored form as any other. Exactly-once is untouched (the `spec_kickoffs` claim, the `HasAny` stand-down, the one-active-turn guard).

**F. Root cause — a reload stranding the interview's answers.**

The room stored the co-edited `answers` **inside** the mirrored question value, so both shared one `Y.Map` key. A reload runs this sequence:

1. a fresh local `Y.Doc` is created — the Hocuspocus provider is still handshaking;
2. the chat log is available immediately (localStorage cache, then the thread bootstrap), so `useRoomQuestion` mirrors the question into that empty local doc;
3. the user answers into it;
4. the room's persisted state — carrying the **pre-reload session's** write to the same key — finally syncs in.

Step 4's write is *concurrent* with steps 2–3, and `Y.Map` settles a tie on one key by **clientID**, which is random per doc. Whenever the pre-reload client's id was the higher one — about half of all reloads — the incoming copy won and took everything in the value with it, `answers` included, so `allAnswered` could never be satisfied. (Verified directly: with `local.clientID = 1` and `persisted.clientID = 2` the new test fails on the old code and passes on the new.)

Fix: the answers moved into their own map, keyed `<toolCallId>#<index>`. The incoming copy has nothing to say about those keys, so there is no tie to lose; per-**index** keys also stop two people answering different questions from overwriting each other. A room written by the old build has its inline draft migrated into the new keys on first mirror, so an interview open across the upgrade keeps what was already selected. `mirrorQuestion` additionally no-ops when nothing changed (the log re-mirrors on every change, and an identical write still woke every observer in the room).

One more thing that class of bug deserves: **the form now names what is blocking Continue** ("1 question still needs an answer"). A free-text question — no options, because the agent wants a typed answer — reads as the optional note field beside the choice cards, so "every radiogroup shows a selection" and "everything is answered" are genuinely not the same thing, and only the form knows the difference. A disabled Continue should never again be a mystery.

**Verification**

```
apps/console: tsc -p tsconfig.json --noEmit → OK
apps/console: eslint . → OK
apps/console: vitest run → Test Files 88 passed (88) | Tests 899 passed (899)
services/aep-api: go build ./... → OK
services/aep-api: go test ./internal/projects/... ./internal/spec/... → ok (all packages)
  --- PASS: TestCreateProject_KicksOffSpecWhenPromptExists
  --- PASS: TestCreateProject_KicksOffSpecWithoutPrompt
  --- PASS: TestCreateProject_KickoffFailureDoesNotFailCreate
gates: make deadcode-ts-check → OK; make -C services/aep-api deadcode-check → OK
```

Not done: no live drive of the deployed build from here, so F's fix is proven by a deterministic unit reproduction of the CRDT race rather than by a second live reload.

hevayo:
Code-review pass on the round-3 commits — three fixed, two left open with reasons.

**Fixed: the answers split was incomplete** (`questionRoom.ts`). `update` hands back a FULL normalized array, so the first edit published an empty answer for every *other* question too — claiming keys the incoming room copy may already hold real answers under, and reopening the exact clientID tie the split exists to remove. Concretely: answer q0 before the reload, answer q1 after, and q0's answer was discarded about half the time — Continue stuck on "1 question still needs an answer". An empty answer for a key that does not exist yet says nothing and is no longer written; clearing an *existing* answer still is. The legacy migration got the same rule. New test pins the case the first reload test missed (it answered everything locally, so there was no per-key conflict to lose), and it fails on the previous commit.

**Fixed: the log went stale whenever the chat rail was closed** (`useChatLogBootstrap.ts`). `useAgentChat` is the log's only live writer and exists only while the rail is open (`Collapse unmountOnExit`). Open the chat during the first run, close it, and the agent's `ask_questions` never reached the log — so never the room, never the shared form; the spec view sat on its working state until the panel was reopened or the page reloaded. That is precisely the chat-rail dependency the bootstrap was added to break, and the one-shot "seed only an empty log" rule left the second half of it in place. The bootstrap now keeps the thread current on the same 12 s cadence whenever no panel owns the log, and stands down the moment one does (ref-counted `registerChatLogOwner`, mirroring the deterministic-flush registry). Both writers project the same server history through `projectableHistory`, whose ids are position-stable, so handing the log back and forth changes nothing on screen.

**Fixed: `KickoffSpec`'s wiring guard missed `turns`** (`kickoff.go`), which the claimed path calls via `HasAny`. A nil there panics on the projects domain's *detached* goroutine — a process crash instead of the logged best-effort failure that domain is promised. Guard extended and `TestKickoffSpec_RefusesWhenUnwired` table-ised over all three stores.

**Left open — the provisioning window** (review findings 4 and 5). Two observations that pull against each other:

- while the repo provisions the kickoff has not started its turn yet, so the thread is empty, `started` is false, and the card still offers "Generate spec" with `?generate` — a click there races the kickoff, and whichever loses the one-active-turn guard surfaces a `TurnInProgress` error;
- but suppressing the CTA in that window would also suppress it after a kickoff that *failed*, and the documented fallback ("a failed kickoff costs the user one click") is the only way back.

Telling those two apart needs a server-side signal for "a kickoff is claimed and pending" that does not exist today — the `spec_kickoffs` claim row is not exposed. I did not want to guess a contract addition at the end of a live-testing round. Happy to take it as a follow-up: expose the claim's state on the project status and let the card render "starting…" instead of either CTA.

hevayo:
Fourth round — the Generate CTA no longer races the BE kickoff: the console does not start the spec interview at all any more.

**Root cause.** The overview's Spec card attached `?generate=requirements`, which `AgentChatPanel`'s autoGenerate effect turned into a composed `/start`. Since #485 the backend fires that turn itself at project create, so the console's send was a *second* turn, and aep-api's one-active-turn guard (D18) rejected it — surfaced verbatim as **"An agent turn is already running for this project — wait for it to finish."**

Two things made it easy to hit rather than rare:

* The panel's only guard was `useAgentEngaged`, which reads the **local chat log**. A fresh tab, a cold cache, or the seconds right after create all read "idle" while a turn streams server-side.
* The card's `open` guard is fed by `useSpecInterview.started` → the active-turn poll, `storeHasLog`, and the thread rehydrate. All three are false for the **2.4–5.7 s between the kickoff claim and the `agent_turns` row** (measured on the live stack). In that window the button reads "Generate spec" and attaches the param; by the time the panel resolves its thread id and sends, the turn row exists. That is why the failure looked like it came from the spec CTA even though the guarded "Continue spec" label never attaches `?generate`.

**Stale `?generate` in the URL: not a real path.** Verified in `@tanstack/router-core@1.171.14` (`buildLocation` → `if (!dest.search) return {}`): a `navigate` without `search` produces an empty search object, so nothing carries a param forward across navigations. `AppLayout.clearGenerate` already strips it on consumption (`replace: true`). The only re-fire vector was a URL whose param was never consumed (a reload or Back onto it) — which the active-turn gate now handles by consuming it and sending nothing.

**What changed**, in three commits:

1. **`f946102` — the requirements injection is gone.** No `?generate=requirements`, no composed `/start`, no `START_COMMAND` in the console. The Spec card is navigation; its label only says where it leads ("Continue spec" / "Open spec"). `?generate=design` stays (design has no backend kickoff) and gains the guard: the send waits for `GET /turns/active` to **answer**, and suppresses itself if a turn is running — the same question aep-api asks when it accepts or rejects a send. A pending chat seed (Resolve-via-chat, submitted interview answers) is now *held* rather than consumed while a turn runs, since a 409 would destroy content that exists nowhere else.
2. **`d1cb62a` — the kickoff records what became of it.** `spec_kickoffs` carries status + reason + updated_at; `GET …/status` exposes `spec.kickoff` as `none | pending | failed(reason) | started` (read only for a project with no spec, so the settled majority pays nothing). `POST /projects/{name}/spec/kickoff` is the retry: one attempt, no provisioning loop, idempotent by state, and a failure comes back **in the body** so the caller renders the same card either way. No automatic retry — an agents service that is down has nothing to retry into.
3. **`96fb7b1` — the card is honest about failure.** pending → "Agent is starting the interview"; failed → the attempt's own reason in error tone plus **Retry**, which asks the backend to run its kickoff again and never sends chat text; the spec view stays one click away throughout.

**What each test pins** (`apps/console` vitest 920 pass, `services/aep-api` `go test ./...` incl. the Docker DB tier):

* `OverviewPipeline.test.tsx` — the card never attaches a `search` in **any** state, and never offers "Generate spec"; the claim window (kickoff `pending`, no turn row, empty chat log) says the interview is starting and offers no way to start another; a `failed` kickoff shows its reason and Retry; Retry calls the backend for **this project** and does not navigate; Retry locks while in flight; a retry that cannot reach the backend is reported separately from the kickoff that failed.
* `AgentChatPanel.test.tsx` — design sends when nothing runs; **sends on a project with prior history** (Generate design is only ever clicked on such projects — history must never suppress it); holds *unconsumed* until the thread id resolves and until the active-turn read answers; sends nothing when the server reports a running turn with an empty local log, and consumes the param so it cannot re-fire; sends again once the read says the turn ended (recovery); a seed is held, not consumed, while a turn runs and goes out when it ends.
* `useActiveTurn.test.tsx` — `resolved` is false until the read answers (gating on an unresolved read is the bug), a failed read resolves as "no turn" so recovery stays possible, and a disabled read issues no request.
* `spec/kickoff_test.go` — pending/started/failed/stalled/none state derivation, including a legacy row with no status; the recorded outcome; reason wording (provisioning cases named, unknown causes carried and capped); `runRetry` re-arms and attempts after a failure, is a **no-op** while started *or* pending, claims when nothing was ever claimed, and returns a still-failing attempt as state.
* `spec/repository_kickoff_dbtest_test.go` (real Postgres) — a fresh claim is `pending`, `SetOutcome` stores status+reason and moves `updated_at`, `Rearm` returns to pending without un-spending the claim, and `SetOutcome` never invents a claim row.
* `projects/project_stages_test.go` — the state reaches `status.spec.kickoff`, is skipped once a spec exists, and reports `none` when the port is unwired.

**Notes / open points**

* A pending claim older than `KickoffWindow` (= `projects.specKickoffTimeout`, one constant now) is *reported* as failed so Retry is reachable after a process death mid-attempt. It is a read-time rule — nothing is rewritten, and no sweep or backfill was added.
* Existing rows gain the columns via AutoMigrate and read as pending → after the window, as retryable. No backfill, per the scope trim; the orphan audit of the live DB was dropped on the same instruction, so I have not queried it.
* `KickoffFailureReason` carries the raw cause (capped at 300 chars) for anything that isn't a known provisioning case. That is deliberate for a platform-engineer console, but it is the one place raw server error text reaches the UI — say the word and I'll reduce it to a fixed sentence plus a log correlation.

hevayo:
Fifth round — copy alignment and real failure detection:

### 1. One voice across the three surfaces

The card narrates **about** the agent; the chat **is** the agent. Both now say the same thing about the same thing — "your idea" — and the count lives in exactly one place.

| surface | before | now |
|---|---|---|
| Spec card, reading | `Agent is processing the idea` | **`Agent is looking at your idea`** |
| Spec card, questions pending | `Agent has 4 questions` | **`Agent has 4 questions about your idea`** |
| Chat, after `/start` | `Looking at your idea to generate the product requirements document…` | **`Looking at your idea…`** |
| Chat, before the banner | `I have a few more questions before generating the PRD.` / `One question before I can generate the PRD.` | **`I have some clarifications about your idea before I write the PRD.`** |
| Spec view, reading | `The agent is processing the idea…` | **`The agent is looking at your idea…`** |

The chat's handoff line drops its count (and with it the singular variant): the banner directly below it and the card both already carry one, and inflecting a third copy was the drift this round set out to remove. Tests that pinned the old strings are updated — `OverviewPipeline`, `TurnBlock`, `AgentChatPanel`, `SpecView`, `SpecQuestionFormReload`.

`89b14dc`… see `copy(console): the card and the chat say one thing about your idea`.

### 2. The spec view never offers a file picker to a project with no files

Opening the spec view before the questions were injected — before the active-turn poll had resolved anything — showed `Select a file to view its content.` on a project that has no files to select. The working state was gated on a **resolved** interview stage, so every moment before one fell through to the file browser's empty state.

The gate is now the project's own condition: **no spec, no files, no failed derivation** — independent of `interview.running`. The stage line became a ladder that always ends somewhere:

1. drafting → `The agent is drafting the PRD…` (unchanged)
2. running → `The agent is looking at your idea…`
3. questions waiting → `Opening the agent's questions…`
4. otherwise → **`The agent is working on your spec. This view fills in as it writes.`**

Same spinner treatment throughout. `Select a file…` is left to the project it is true of: one whose spec exists.

### 3. Failure detection now keys off the turn's outcome, not the dispatch

You were right that it did not fire for the case that matters. The fourth round's path only covered a dispatch that never produced a turn. Your `docker stop aep-agents` run produced the opposite shape, and it is still sitting in the dev DB:

```
spec_kickoffs
 org_id  |     project_id      | status  | reason
 default | pantry-tracker-what | started |

agent_turns  (the project's only turn)
 status | reason          | message
 failed | dispatch-failed | agents dispatch failed: agents service turn request:
        |                 | Post "http://agents:4000/conversations/…/turns":
        |                 | dial tcp: lookup agents on 127.0.0.11:53: server misbehaving
```

`aep-api` creates the turn and stamps the claim **synchronously**; the turn then runs detached and fails seconds later, long after anything is left to record it. So the claim says `started` and always will — and the old read treated *any* turn row as success on top of that.

**What now flips the card to failed** — all three, together:

- the project has **no spec** (`populateStages` only asks for the kickoff state when `!snap.HasSpec`, unchanged), **and**
- **no turn is running and none has ever completed**, **and**
- **at least one turn ended `failed`** — the newest one's `message` (or its failure class) becomes the card's reason.

`TurnRepository.HasAny` is replaced by `Standing`, which returns exactly that: `Progressed` (running-or-completed) plus `LastFailure` when nothing has. A running or completed turn always outranks an earlier failure, so a first run parked on its questions, and a Retry whose new turn is streaming, both read `started`.

The kickoff's **stand-down** moves to the same question. It used to ask "does a turn row exist", which the corpse answered yes to — so **Retry stood down on the failed attempt, reported success, recorded `started`, and started nothing**. It now stands down only on a turn that progressed. Retry stays idempotent and user-triggered; there is still no automatic retry, no backfill and no sweep — both failure reads are derived at read time and never written back, so a Retry's real outcome still wins.

### Verification

**DB tier, against Postgres, through the real repositories** (`internal/spec/repository_kickoff_dbtest_test.go`) — the state transition itself, not fakes:

```
--- PASS: TestTurnRepo_StandingSeparatesALiveTurnFromADeadOne (2.75s)
--- PASS: TestKickoff_ADeadFirstRunTurnReadsAsAFailedKickoff (2.75s)
PASS
ok  	github.com/wso2/aep/aep-api/internal/spec	3.116s
```

`TestKickoff_ADeadFirstRunTurnReadsAsAFailedKickoff` walks the whole sequence: claim taken → turn started → outcome recorded `started` → **`Kickoff` says `started`** → turn `Finish`ed as `failed`/`dispatch-failed` → **`Kickoff` says `failed` carrying the turn's cause** → claim row asserted still `started` (derived, not written back) → retry turn attaches → **`Kickoff` says `started`** again.

**Against your actual stuck project.** I built this branch's `aep-api` and ran the fixed `Kickoff()` read-only against the live dev Postgres:

```
pantry-tracker-what      -> status="failed" reason="The spec interview's first turn failed: agents dispatch
                            failed: agents service turn request: Post \"http://agents:4000/conversations/
                            org_default--proj_pantry-tracker-what--general--4ea9…/turns\": dial tcp: lookup
                            agents on 127.0.0.11:53: server misbehaving"
gym-workout-tracker-dfg  -> status="started" reason=""
```

The project your agents-down run left stuck now reports failed with the real dial error, and a healthy neighbour is unaffected.

**Honest gap:** I could **not** drive the agents-down path end to end through the UI — the console sign-in needs your credentials (the seeded `mark`/`emily` users are not in this Thunder instance), so I could not create a fresh project with `aep-agents` stopped and watch the card myself. What I verified is the state transition against real Postgres and the fixed read against the real stuck row; the card's failed+Retry rendering itself is the fourth round's code, unchanged. **Both round-5 images are deployed to the local stack** (`aep-api` and `console`, previous images kept as `:pre-round5`), so `pantry-tracker-what` should show the error and the Retry on your next look.

**Rest of the gates:** console `921 passed (89 files)`, `tsc --noEmit` clean, eslint clean on the three touched features, `knip --production` clean; `go test -short ./internal/spec/... ./internal/projects/...` ok, `go vet ./...` clean, `golangci-lint run ./internal/spec/...` 0 issues, `make deadcode-check` OK.

No contract change — the `failed` enum and the Retry endpoint already existed; only what fills them changed.

PR #525, opened 2026-08-18
fix(scripts): project-rebuild deploys the build it just made

KCSAbeywickrama:
## Problem

ADR-0017 flipped components to `autoDeploy: false` and moved the deploy into the run supervisor's cycle stage. `scripts/project-rebuild.sh` predates that and was never updated, so its last third became wrong:

- It builds the image, the build posts a Workload, and **nothing promotes it** — the only caller of `DeploymentService.Deploy` is `delivery/run/activities.go:443`, and the BFF exposes no deploy trigger (`/deployments` is GET-only).
- The "Waiting for AutoDeploy to move the ReleaseBinding…" block therefore always times out and prints *"the rollout is still in flight"*, which is false.
- Its header comment — *"There is no deploy call to make afterwards, and none exists"* — is now true for the wrong reason.

The hand-edit loop the script exists for (clone repo → edit → push → rebuild) stopped being end-to-end.

## What changed

The script now performs the same two writes the run supervisor's deploy stage performs (`projects.DeploymentService.deployOne`), against the OpenChoreo API:

1. **Cut the release** — `POST /api/v1/namespaces/{org}/components/{project}-{component}/generate-release`, named `{project:18}-{component:18}-{millis}` off the build run's unix-milli suffix.
2. **Re-pin** — read-modify-write of `spec.releaseName` **alone** on the existing binding, retried 3× on 5xx (OpenChoreo reports a lost write race as a generic 500, same reason the Go client wraps this in `retryStaleWrite`).
3. **Wait honestly** — poll the binding's Ready condition, gated on `observedGeneration` advancing past what the pin replaced.

Supporting changes:

- Pre-flight resolves the OC token, the org and the binding **before** spending a build, and refuses a component with no binding — a pin-only write would create one carrying no traits, env vars or files, the case `DeploymentService.Converge` refuses for the same reason.
- Uses the BFF's own service identity (`aep-api-client`) for OpenChoreo, not the seeder client: OpenChoreo authorises per resource and the seeder token is **403 on `releasebindings`**. Both credentials are `deployments/docker-compose.yml`'s local-dev defaults and are overridable (`OC_CLIENT_ID`/`OC_CLIENT_SECRET`).
- Exit code now reflects the deploy, not just the build.

### Why the release name is build-derived, not commit-derived

`EnsureRelease` treats 409 as success (`component_chain_client.go:103`). A commit-derived name would 409 onto the existing release whenever you rebuild without a new push (a Dockerfile fix, a retry) and **silently keep serving the old image**. The build run name carries a unix-milli suffix (`oc_names.go:59`), so it is unique per build and stable across a re-run of the script for the same build.

### What it deliberately does not do

The deploy stage rewrites the **whole** binding from the current design in one call — release pin, trait environment configs (jwtAuth) and workload overrides (env vars, `env-config.js`) — via `DesiredDeploymentFor`. This writes only the pin, which is exactly why the wiring already on the binding survives untouched. So a hand-edit to **code** deploys as the platform would; a change to the **design** does not reach the cluster through here. Documented in the script header.

It also deploys one component, skipping the wave ordering a run derives from the design's hard wiring edges (ADR-0019) — safe in this loop because the component's dependencies are already up.

## Verification

All against a live local plane. **Read-only — no release was cut and no binding was written.**

```
1. seeder token: ok
2. oc token: ok
3. org resolved: 'default'
4. names: scoped='p18-bare-minimum-hello-hello-webapp' binding='p18-bare-minimum-hello-hello-webapp-development'
5. binding pre-check: HTTP 200
6. missing-binding branch: HTTP 404
7. release name: 'p18-bare-minimum-h-hello-webapp-1755400000000' (len 45, budget 63)
9. rmw preserves siblings: {"releaseName":"...","kept":["environment","owner","releaseName","state","workloadOverrides"]}
```

`generate-release` authorises correctly — a fake component returns `404 {"code":"NOT_FOUND","error":"Component not found"}`, i.e. past auth, not rejected by it.

**The endpoint URL does not move on a re-pin.** Every name in the serving path is keyed on project/component/environment, never on the release. Empirically on `p18`: the RenderedRelease was created `2026-08-03T11:51:49Z` and the HTTPRoute is 14d old, while the binding pins a release created `2026-08-14T05:38:52Z` — re-pinned 11 days later, both objects updated in place with identity and timestamps intact.

### A bug caught and fixed during verification

The first version of the readiness gate compared the Ready condition's `observedGeneration` against `metadata.generation`. **openchoreo-api strips `generation` from metadata** (kubectl reports `4`; the API response has no such field), so the gate was inert and would have reported success instantly off the *previous* generation's `Ready=True` — precisely the lie it existed to prevent. It now requires the stamp to advance past what the pin replaced, with a no-op path when the pin changes nothing. Gate truth table, exercised under bash 3.2:

```
E. real re-pin, stale Ready=True from old gen : WAIT (not observed yet)   <- the bug
F. real re-pin, controller caught up          : READY
G. real re-pin, caught up but not ready       : POLL
H. no-op re-pin (same release, no bump)       : READY
I. no-op re-pin, not ready                    : POLL
```

Field parsing uses `|` rather than tabs because `read` collapses runs of whitespace separators, which shifted fields whenever a condition reason was empty. Verified with empty `status`+`reason` (obs still parses as `7`) and with a conditions-less binding (no crash, obs `0`).

`bash -n` clean. Local bash is 3.2.57, so no 4.x-only parameter expansions are used.

## Not yet verified

**The two write calls and the full build→deploy loop have not been run end-to-end.** That needs a real build (minutes, an image push) and a live re-pin on a test project. Everything above is read-only probing of the same paths, payload shapes and decision logic. Happy to run it against a scratch project and attach the output before merge.

`/code-review` has not been run on this yet either.

## Note for reviewers: existing components still carry `autoDeploy: true`

Every user component on my local cluster (`p12`, `p15`, `p16`, `p18`, `p24`, `p25`) still has `autoDeploy: true` — `EnsureComponent` only re-asserts `false` when a real cycle next touches them. Until then OpenChoreo promotes the build's Workload itself and races this script over one field. Two guards were added rather than mutating the flag:

- a pre-flight **warning** with the `kubectl patch` to settle it, before minutes are spent on a build;
- the wait loop re-reads `spec.releaseName` each poll and **fails loudly** with that diagnosis if the pin is taken from under it.

The script does not flip the flag itself — that is a platform invariant `EnsureComponent` owns, and a script that leaves it flipped would hand a later run two writers on one binding's pin.

Also worth flagging as pre-existing staleness this PR does not touch: `component_service.go:39-56` and `componentbuild/handler.go:86` still document the AutoDeploy-drives-deploy chain.

### DIFF
--- a/scripts/project-rebuild.sh
+++ b/scripts/project-rebuild.sh
@@ -15,9 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# Rebuild one component of a platform-created project from what is on its
-# default branch — the local-dev trigger for
-# POST /projects/{project}/components/{component}/builds.
+# Rebuild AND deploy one component of a platform-created project from what is on
+# its default branch.
 #
 # The loop it was written for: clone the repo the platform created for a project,
 # change code by hand, push to the default branch, run this. The push alone does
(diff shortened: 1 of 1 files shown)


KCSAbeywickrama:
After discussion, we decided to have a feature to auto-build and redeploy when a PR is merged

PR #549, opened 2026-08-19
fix(console): surface gate-refusal details when Build fails from the dependency drawer

senithkay:
## Summary

- When a build fails the spec validation gate and the user went through the dependency drawer (has external dependencies), `onContinueBuild` only showed the generic `"spec validation failed"` message with no per-file detail
- The `runBuild` path (no dependencies, Cut-version dialog) already extracted `e.details` and rendered the actionable gate checklist — this aligns the drawer path with the same behaviour
- The drawer is also closed before the checklist is shown so the overlay doesn't occlude it

## Changes

**`SpecView.tsx` — `onContinueBuild`:**
- Extract `e.details` in the catch block (same cast as `runBuild`)
- If details are present: close the drawer, call `setGateRefusal(details)` to render the per-file checklist
- If no details: fall back to `setBuildError(message)` as before
- Clear `gateRefusal` on entry (mirrors `runBuild`)

**`SpecView.test.tsx`:**
- New test: drawer Continue with a gate refusal — closes the drawer and surfaces the gate checklist
- New test: drawer Continue with a non-gate error — surfaces the plain error (regression guard)

## Root cause

`onContinueBuild` and `runBuild` both call the same build mutation but handled the error differently. The `runBuild` path checked for `e.details` and called `setGateRefusal`; `onContinueBuild` did not. Commit `10c8f535` added the gate checklist UI to `runBuild` but left the drawer path unpatched.

## Test plan

- [x] `SpecView.test.tsx` — all 23 tests pass, including the two new ones
- [ ] Manual: deploy with dependencies, click Build, fill in secrets, click Continue — should now show the "Build refused — the design isn't complete" checklist with per-file items instead of the generic error


senithkay:
Closing — branch still carried unrelated aectl changes. Replacing with a clean 2-file PR.

PR #596, opened 2026-08-22
fix(aep-validation): a validation run could load no skills, moved its own shell, and never pushed

KCSAbeywickrama:
Three defects in the validation run's own instructions, each of which cost real tokens on every run and none of which failed loudly enough to notice. All three are fixed and measured against real runs against `ChamodOrg/p26-bare-minimum-hello`.

## 1. A validation run could not load *any* skill on demand

`oneshot.ts` populated `availableSkillNames` only in the coding branch, so a validation run passed `skills: []` to the SDK. Since `skills:` is an **allowlist over discovered skills**, `[]` allows nothing — every `Skill` invocation in a validation run was rejected, not just `playwright-cli`:

```
Skill playwright-cli is not in this session's skills allowlist
```

**How it hid for ~3 weeks.** `aep` and `aep-validation` reach the model as system-prompt text via `alwaysOnSkills` + `requireWorkflowBodies`, so the run looked healthy — `playwright-cli` is the only skill a validation run must reach *through the Skill tool*. And the guard built for exactly this, `checkPreload(skills, message.skills)`, compares *requested* against *resolved*: with `[]` requested, `missing` is always empty, so it can never fire.

**What it cost.** The agent compensated by reconstructing the skill by hand — 20 consecutive tool calls grepping `SKILL.md` out of the filesystem and re-deriving DNS/host-resolver behaviour the config template already encodes:

```
seq 56  which playwright-cli; npm ls -g; find / -iname "*playwright*"
seq 58  ls -la /app/skills/playwright-cli; playwright-cli --help
seq 62  cat /etc/hosts; getent hosts …
seq 66  node -e "require('dns').lookup(…)"
seq 70  grep -n -i "host-resolver|proxy" /app/skills/playwright-cli/SKILL.md
seq 72  sed -n '190,260p' /app/skills/playwright-cli/SKILL.md
seq 92  grep -n "config" /app/skills/playwright-cli/references/*.md
seq 94  grep -rn "host-resolver-rules" …/@playwright/cli
```

The run still reported success. A green run built on guesswork is the worst shape of this bug.

**Fix.** `onDemandSkills(taskKind)` beside `alwaysOnSkills`, returning `["playwright-cli"]` for validation and `[]` otherwise, resolved against the mirror so an absent skill is named at startup instead of arriving as a rejected call mid-run. Deliberately one name rather than the whole mirror — a validation run builds nothing, and `runner.ts` already argues against `"all"`.

**Verified.** Post-fix runs show the new startup line and a successful load:

```
seq  8  [oneshot] 1 skill(s) loadable on demand: playwright-cli
seq 38  Skill playwright-cli  →  ok: true
seq 40  playwright-cli open / snapshot / eval / close   ← against the live app
```

## 2. The workflow moved the shell, then assumed it hadn't

`SKILL.md`'s RUN block opened with a relative `cd tests/e2e`. The Bash tool keeps **one shell for the whole run**, so that `cd` persisted — and the second time the block ran, it resolved against `…/tests/e2e`:

```
/bin/bash: line 1: cd: tests/e2e: No such file or directory
```

Underneath sat a contradiction the skill carried silently: `healing.md`'s re-runs were bare (`npx playwright test`) and worked *only because* the `cd` had persisted, while the REPORT step and `generate-report.mjs`'s defaults are all repo-root relative. HEAL depended on the shell having moved; REPORT depended on it not having.

**Two failed attempts before the fix that worked**, both recorded here because the negative result is the useful part:

| attempt | kind | outcome |
|---|---|---|
| self-locating `cd "$(git rev-parse --show-toplevel)/tests/e2e"` | snippet | **ignored** in both runs that had it |
| "one shell serves the whole run" as a stated fact | statement | superseded before it could be isolated |
| remove the `cd` entirely | **structural** | **worked first try, completely** |

The agent composes its own commands rather than copying the skill's — note the `&&` chaining and `| tail -60`, neither of which is in the skill:

```
seq 48  cd tests/e2e && npx playwright-cli open …      ✗ cd fails
seq 50  pwd                                            ← wasted recovery
seq 52  npx playwright-cli open …                      ← retry, works
seq 54  cd tests/e2e && npx playwright-cli snapshot    ✗ fails AGAIN
```

It re-prefixed `cd tests/e2e` at seq 54 *after* already being burned at seq 49. It does not learn within a session, which is why instructing it better was never going to hold.

**Fix.** One line in the scaffolded `package.json`:

```json
"scripts": { "test": "playwright test" },
```

`npm --prefix` runs a script with the package as its working directory, so Playwright finds its config without the shell ever moving. Every invocation became `npm test --prefix tests/e2e`, filters ride a `--`, install is `npm install/ci --prefix tests/e2e`. **Zero `cd` now remains in `SKILL.md`, `authoring.md` or `healing.md`.**

**The rejected alternative, and why.** `playwright test --config tests/e2e/playwright.config.ts` from the root was tested and **breaks the report silently**: it reports spec paths relative to the repo root, and all three of `generate-report.mjs`'s resolution candidates assume they are relative to `tests/e2e`, so the spec column degrades to `tests/e2e/tests/e2e/specs/…`. The package-script route keeps reporter output byte-identical, which is why the generator needed no change at all.

Verified against the pinned Playwright 1.61.1 with the real template, every command from the repo root:

```
npm test --prefix tests/e2e                     → results.json in tests/e2e/test-results/
npm test --prefix tests/e2e -- specs/X.spec.ts  → filter works
generate-report.mjs --issue 3 --commit …         → exit 0, "e2e 1/1 passing"
  report.json  spec: tests/e2e/specs/AC-001-a.spec.ts     ← the join option A would have broken
```

The `// spec:` header contract check passing proves the generator *resolved and read* the file, not merely printed a path.

## 3. The push was never documented, and it always has to force

Step 10 was `gh pr create` alone — no `git push` anywhere in the workflow. So the agent invented one, and it was rejected:

```
seq 137  error: failed to push some refs
seq 140  git push --force-with-lease            ← recovered on its own
```

This happens on **every** re-validation, because three facts compose: the branch name carries no cycle discriminator (`aep/m<N>-validation`, unlike coding's `aep/m<N>-c<k>` counted from existing remotes), the platform **squash-merges** so a merged cycle's tip is no ancestor of `main`, and nothing deletes the branch. Confirmed in the test project: `aep/m1-validation` still carries three merged PRs, one per re-validation.

The deny-list made it worse — it summarised the `aep` rules as a flat *"no force-push"* while `aep/SKILL.md` actually permits one on the run's own branch with `--force-with-lease`. The summary was **stricter than the list it summarised**, so the agent reached the right outcome by disobeying its own skill.

**Fix.** Document the push, unconditionally forced, and correct the deny-list. Verified on a scratch repo that one command covers both cases:

| | plain push | `--force-with-lease` |
|---|---|---|
| branch doesn't exist yet | `* [new branch]` | `* [new branch]` |
| re-validation (divergent) | `! [rejected] (non-fast-forward)` | `+ 39928be…6d5a817 (forced update)` |

Keeping the deterministic name over adding a `-c<k>` discriminator accepts one consequence: force-pushing over an unmerged crashed cycle discards its commits. There is nothing to resume — step 3 re-derives coverage from what is committed on `main`, which is why validation needs none of the crash-resume machinery the coding skill carries.

## Measured result

Five runs on the same project, same criterion, same authoring path, **same runner image** (the only source delta between the two `cd` arms is a test file that never executes at runtime).

| run | calls | `cd` | fails | **cd-fails** | out tok | cache read |
|---|---|---|---|---|---|---|
| #6 baseline (allowlist bug live) | 67 | 26 | 3 | 1 | 23,066 | 4.04M |
| #9 with `cd` | 27 | 13 | 1 | 0 | 7,217 | 1.49M |
| #11 **arm A**, with `cd` | 37 | 12 | 3 | **2** | 9,955 | 2.16M |
| #10 no `cd` | 24 | **0** | 1 | 0 | 6,057 | 1.42M |
| #12 **arm B**, no `cd` | 31 | **0** | **0** | 0 | 9,717 | 2.17M |

```
with cd   mean of 3:  calls 43.7   out tok 13,413   cd-failures 3
no cd     mean of 2:  calls 27.5   out tok  7,887   cd-failures 0
```

**What is solid:**

- **`cd` calls separate absolutely** — 26 / 13 / 12 with, 0 / 0 without. No overlap. Structural, not sampled: a workflow that names no working directory cannot produce one, and the failure that follows is unreachable at zero.
- **`cd` failures: 3 across 3 with-`cd` runs, 0 across 2 without.** Arm A cost four calls to it in a single run.
- **The allowlist fix removed the 20-call detour** and, with it, an entire class of guessed-at browser mechanics.
- **Arm B is the only run of the five with zero failures.**

**What is not solid, stated plainly:**

- n=3 vs n=2. The mean difference is suggestive, not established.
- Within-arm spread is roughly ±30% (27 vs 37 calls with `cd`; 24 vs 31 without), driven by route choice — #12 explored with `playwright-cli` where #10 wrote the spec source-informed from `App.tsx`. Both are legitimate per `authoring.md`. So treat the token figures as "meaningfully cheaper, magnitude uncertain" rather than a percentage.
- **Wall time is not an independent metric** — per-call latency is flat at 5.6–6.2s, so wall time is call count in disguise. Arm B took *longer* than arm A while failing less.
- Absolute-path usage swings 24 / 14 / 2 within the with-`cd` arm alone; it is noise, not signal.

Every result above was confirmed against the mirror the run actually read, at the run's own commit. That check is not optional: one earlier run was credited to a change it had never received (see #605).

## Guards

Doc changes can't be unit-tested, so each fix is pinned by a content guard beside the existing ones in `workflow_skill.test.ts`, and each was proven by mutation rather than merely left green:

| guard | catches |
|---|---|
| `the validation workflow never runs playwright test bare` | a bare `npx playwright test` returning, and the `test` script being dropped from the template |
| `the validation workflow never cds to a bare relative path` | a relative `cd` reappearing (passes vacuously now — the state we want pinned) |
| `no skill licenses a force-push without the lease` | library-wide, any `push` line with `--force` lacking `--force-with-lease` |
| `aep-validation still names the force-push its push step needs` | the deny-list outliving the step that requires the force |
| `onDemandSkills: a validation run may load playwright-cli` | the allowlist regressing, paired with the existing assertion that it is *not* always-on |

The bare-invocation guard earns its place: from the repo root a bare `npx playwright test` **discovers the specs, prints `1 passed`, exits 0** — and loads no config, so no reporter, no `results.json`, and none of the launch args the endpoints need. The new way to get this wrong is quieter than the old one.

**397/397 tests pass, tsc clean.**

## Scope

Confined to the validation surface: `skills/aep-validation/` and `runners/remote-worker/`. **`skills/aep/` is untouched** — verified branch-wide — because it steers every coding run of every project. One consequence is accepted and worth stating: the shell-persistence fact is not given to coding runs. The coding workflow has no `cd` of its own today, so there is nothing there to trip on now.

A rename rides along: `resolvePinnedSkills` → `resolveSkillPresence`, `PinnedSkillsResolution` → `SkillPresence`, `.preload` → `.present`. The field was the worst of it — `availableSkillNames = preload` reproduced the exact pinned-vs-allowed conflation this PR is about, inside the code fixing it.

## Follow-ups filed

- #604 — a first validation's probe truncates itself on the absent `tests/`; the structural fix is Go-side (spec inventory in `validation-context.json`). Includes the reverted attempt and why instructing the agent did not work.
- #605 — `build-runner.sh` and `skills/AGENTS.md` both claim a skill edit needs no rebuild, citing an overlay nothing consumes. This cost a wasted run.

## Proof of execution

All five runs are real dispatches against `ChamodOrg/p26-bare-minimum-hello`, each landing a merged PR with committed specs and a generated report. Pod NDJSON logs were read for every one; the sequence numbers quoted throughout are from those logs.


sachiniSam (approved):

KCSAbeywickrama:
@coderabbitai review

KCSAbeywickrama:
@coderabbitai review this even it is closed. The comments will addres in diffrent PR

PR #633, opened 2026-08-26
docs(skills): a sibling's wiring comes from the design, not from a comment

xlight05:
## The problem

`employees-submit-expense131` shipped this in `expense-webapp/workload.yaml`:

```yaml
dependencies:
  endpoints:
    - component: expense-api          # ← the friendly name
```

OpenChoreo resolves an endpoint dependency by the **scoped** name, so this matched no
binding. The release rendered, the pod ran, the app served — and the only symptom was a
ReleaseBinding parked at `Ready=False / ConnectionsPending`, with the project reporting
"deploying" for ever. The wiring-conformance check caught it and minted
[issue #6](https://github.com/asdlc-repos/employees-submit-expense131/issues/6), which
cost a **whole coding cycle** (#6 → #7 → PR #8) to change one string.

The correct value was already in the agent's own tree, byte-exact, committed by the
platform before the agent ever ran:

```jsonc
// specs/design/components/expense-webapp/design.json
"wiring": { "endpoint": { "component": "employees-submit-expense131-expense-api", … } }
```

## Root cause: doc drift against a fix that already landed

`183e67a8` (2026-08-01) moved a `component` dependency's endpoint wiring off the live
endpoint catalog and into `design.json` — precisely because the catalog only lists
components that have already **deployed**, so on a first delivery it answers nothing.
Its own commit message names this exact failure.

The agent-facing docs were never moved with it:

| Where | Said | Actually |
|---|---|---|
| `workload-and-wiring.md` kind table | `component` → *"not derivable from `specs/` — the skill body"* | derived, and stamped into `design.json` since 2026-08-01 |
| `SKILL.md` → *The `endpoints:` half* | read the **"Platform-resolved dependencies"** comment | that comment carries `org-service` only |
| `ADR-0013` | *"a sibling's endpoint name comes from a `workload.yaml` nobody has written yet"* | written one day before `183e67a8` superseded it |

So the agent read the table, went looking for a comment, found none — **no
"Platform-resolved dependencies" comment exists on any issue in that repo**, I checked
all of them — and fell back to the only name it had ever seen.

The instruction that was supposed to save it (`# the platform's name — never "correct"
it`, in the template) never fired, because the agent had already been told the value
was not in `specs/`.

## The change

Move the docs to where the data is.

**`references/workload-and-wiring.md`**

- The kind table routes `component` to *its `wiring` object*, like the other two derived
  kinds.
- A new shape table maps the two `wiring` variants onto the two halves of
  `dependencies:` — an `endpoint` object → `endpoints[]`, `ref` + `envBindings` →
  `resources[]` — so *which half* is read off the data rather than off the kind.
- Copy-verbatim now covers both variants, and the project prefix is stated as the
  platform's own name for the component, with the silent failure spelled out:

  > **A `wiring.endpoint`'s `component` carries the project as a prefix** —
  > `<project>-<component>`. It deliberately does not match the name the rest of the
  > tree calls that component […] Any other spelling parses, builds, deploys and serves
  > with the address env var silently absent.

**`SKILL.md`** — the `endpoints:` half resolves a sibling from `design.json`, and keeps
the comment for `org-service` alone. It closes the hole the agent actually fell into:

> A **sibling** (`kind: component`) is already resolved in your own tree […] **That
> holds whether or not the comment below exists.**

**`ADR-0013`** — restated at its current state: `component` sits on the derived side,
and the conformance net checks endpoint targets as well as refs.

**`overlays/local.md`** — the anchor moves in lockstep.

## Gates

- `runners/remote-worker` — **398/398 pass**, on this branch's base. These render
  `SKILL.md` through `local.md`, and `skill_overlay` throws on a stale anchor
  (`replace-text matched 0 times`), so the overlay edit is proven, not assumed.
- `make license-check` — exit 0.

## Not in this PR

The structural fix. `dependencies:` in `workload.yaml` is **100% platform-derived** —
`ScopedComponentName`, the sibling's own endpoint name, `visibility: project`,
`ServiceURLEnvName`. Asking a model to transcribe a value the platform calculated is a
variance source that no wording removes, and the friendly name is the natural thing to
write, so the wording is fighting the grain.

`checkWiringConformance` already holds, at the same moment and in the same process, the
shipped file, the declared truth, and the exact missing values — and it writes an
*issue*. It could write the *file*: reconcile `dependencies:` from `design.json`, commit,
then build off that commit. No PR, no cycle, no model. The precedent exists —
`DerivePlatformResourceFactsAtHead` already commits to `main` outside a PR at the same
pre-tag step.

That is a separate change. This one stops the bleeding for the next project.

https://claude.ai/code/session_01JvanuzxvZmpfvnh2Pbr2wD

### DIFF
--- a/docs/decisions/ADR-0013-derived-wiring-lives-in-the-design.md
+++ b/docs/decisions/ADR-0013-derived-wiring-lives-in-the-design.md
@@ -24,17 +24,21 @@ for information it already had at design time.
 Split the block by what is DERIVED from what is RESOLVED, and give each the
 channel its nature allows.
 
-**Derived — stamped into `design.json`.** Every `platform-resource` and `external`
-dependency carries a platform-stamped `wiring: {ref, envBindings}` object,
-shape-identical to one `dependencies.resources[]` entry so the agent copies rather
-than transforms it. It is derived in the design-save pass (the pre-tag step on
-`POST /build`), committed to the design the agent reads as its spec, and captured
-in the version tag. There is no ordering, no audience, and no idempotency marker
-left to get wrong.
+**Derived — stamped into `design.json`.** Every `platform-resource`, `external`
+and `component` dependency carries a platform-stamped `wiring` object,
+shape-identical to the entry it belongs in so the agent copies rather than
+transforms it: `{ref, envBindings}` for one `dependencies.resources[]` entry, and
+`{endpoint}` for one `dependencies.endpoints[]` entry. A **sibling**'s endpoint
+qualifies for the same reason a resource's ref does — scoped provider name, the
+sibling's own endpoint name, `project` visibility and one `address` binding are
+all pure functions of the design. It is derived in the design-save pass (the
+pre-tag step on `POST /build`), committed to the design the agent reads as its
+spec, and captured in the version tag. There is no ordering, no audience, and no
+idempotency marker left to get wrong.
 
-**Resolved — pushed at cycle dispatch.** The `endpoints:` half genuinely needs
-live resolution (a cross-project provider may not have published; a sibling's
-endpoint name comes from a `workload.yaml` nobody has written yet), so it stays a
+**Resolved — pushed at cycle dispatch.** What is left on the `endpoints:` half is
+an `org-service`, which genuinely needs live resolution because its provider
+belongs to another project and may not have published yet, so it stays a
 comment — but triggered at dispatch, not at gate resolution. Dispatch is provably
 correct rather than lucky: the dispatch predicate already requires that no gate is
 open in the milestone, so everything resolvable has resolved, and the working set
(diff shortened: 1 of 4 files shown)


xlight05:
Folded into #632 (commit `c52baa80`, cherry-picked unchanged) — #632 now carries all four commits and matches what is deployed on the local plane exactly.

Leaving this open rather than closing it myself, since the review threads here are still readable context. Close it as superseded whenever suits; nothing is lost.

xlight05:
Superseded by #632, which now carries this commit as `c52baa80` (cherry-picked unchanged) alongside the gateway-routing work and the review fixes. The two changes are the same class of defect in the same files, so they review better together than apart.

Closing rather than merging. Nothing is lost: #632's description carries the full root-cause writeup from here, and the change is already live on the local plane (org skill library md5-matches the built commit; pending skill updates `count: 0`).

Issue #580, opened 2026-08-20
A console skill carries the console's narration rules


hevayo:
## Problem

The agent talks to a console user as if they were standing in a terminal inside the repo. Verbatim
from the 2026-08-18 walkthrough:

> Full design is under `specs/design/` (design.cell, design.md, security.md, per-component
> design.json/openapi.yaml/wireframes.dsl) with validation criteria at
> `specs/validation/validation-criteria.json`

> Next step: review the PRD, then run `/design`

— the second said while a **Generate design** button was on screen. Two vocabularies for one
action, and the prose points at the one the UI does not offer.

This is not the agent going off-script; the skills **instruct** it:

- `skills/start/SKILL.md:97` — *"point the user at the next step: review `specs/requirements/prd.md`,
  then run `/design`"*

and the system prompt licenses per-flow narration outright
(`services/agents/src/agents/main/prompt.ts`): *"A LOADED skill may define the narration for its own
flow."*

**It becomes urgent because of a decision this map already made.**
[#523](https://github.com/wso2/labs-agentic-engineer/issues/523) renames the artifacts in the UI —
`prd.md` becomes *Product requirements*, `validation-criteria.json` becomes *Acceptance criteria*.
Once that ships, an agent quoting paths is not merely unfriendly, it names things that are no
longer on screen.

## Users

**Developer/User** — every turn, on every surface. Also the **local CLI user**, who is the reason
this cannot be fixed by stripping the paths out of the shared skills.

## Experience walkthrough

1. **A `console` skill carries the console's narration rules.** The shared flow skills
   (`start`, `design`, `amend`, `/settle`, …) stay **byte-identical everywhere**.
2. **The caller includes it for console turns and omits it for a local run** — `CONTEXT.md`:
   *"The **caller** passes the candidate Skills … in the turn request payload."* No new machinery.
3. **It is inlined into every turn's system prompt**, not offered in the catalog — following the
   `organization` precedent at `prompt.ts:77-83`: *"An agent that has to remember to load it asks
   questions the org already answered."* An agent that has to remember its narration rules will
   forget one and quote a path. Being system-prompt-stable also keeps it cache-friendly, and its
   absence leaves the prompt byte-identical — the same graceful-absence property `organization`
   already has.
4. **It outranks per-flow narration.** A rule any skill can opt out of is not a rule, and this one
   is load-bearing for decisions already committed.
5. **The rules, in full:**
[...]

Issue #575, opened 2026-08-20
Spec rail: make the left rail the flow


hevayo:
## Problem

The spec view never answers *where am I*, and its left rail — the one surface a user reads
throughout the whole journey — is inert.

- **Sections are dead headings.** `REQUIREMENTS` / `DESIGNS` / `VALIDATION` never say which is
  live. Two of them read *"Being derived…"* **before the user has asked for a design**, stating
  something untrue about what the platform is doing.
- **Artifacts are filenames.** `prd.md`, `design.cell`, `security.md` — the user is reading a
  repo, not a document tree.
- **Nothing says how far along a run is.** A design turn ran ~90s in the 2026-08-18 walkthrough
  with no sense of what was left. The rail is a log of what appeared, never a plan.
- **Build lives somewhere else** — a header button that appears once a design exists — so the
  sequence the rail implies stops at Validation and the user infers the rest.

## Users

**Developer/User** — in the spec view for most of the journey, which is where the agent does most
of its work.

## Experience walkthrough

1. **The rail is the flow.** Sections in order — **Requirements · Design · Validation** — each
   carrying state: **ready** (green tick) · **active** (pulse, name in primary) · **outdated**
   (amber) · **not started** (dim).
2. **Artifacts are titled, not filenamed**: `REQUIREMENTS › Product requirements`,
   `DESIGN › Architecture · Design overview · Security` then per-component,
   `VALIDATION › Acceptance criteria`. **Repo paths do not change** — the lexicon holds the
   mapping.
3. **Artifacts carry state too**: planned (ghost) · writing/modifying · done · error.
4. **Work in progress is the app's own pulse** — `agentChatWorkingPulse` from
   `WorkingIndicator.tsx`, 8px `primary.main`, not a new animation and not a spinner.
5. **Build sits at the foot of the rail** as its terminal step: *"after the design is written"*
   before, and when the design is stale, **blocked** — *"The design is behind your requirements"*
   with **Update the design**.
6. **"An agent is working" stops being a gate tooltip.** The rail shows it, larger and better.

## Scope

**In:** section state and its four values; artifact titles; artifact state; the pulse; Build as the
rail's terminal step and its blocked state; retiring the *"Being derived…"* placeholder for
**"Not created yet"**; removing the agent-busy tooltip from Build.

**Out (explicitly):**
- **How the plan is declared** — the mechanism that populates *planned* artifacts is the next issue.
[...]

PR #641, opened 2026-08-28
feat(console): the turn declares its plan, and the rail counts it

### LINKED ISSUE #576: Skills declare what they are about to write
## Problem

Progress in the console is **emergent**: the UI can only render a tool call after it has fired, so
it can say what has happened and never how much is left. A design run took ~90s in the 2026-08-18
walkthrough with no denominator anywhere.

The information exists. **The skills know their own plans** — `design` walks a known artifact set
(the cell, then design.md, security.md, then per-component files, then criteria); `start` walks the
PRD's own sections. That knowledge is simply never emitted, so the rail
([#575](https://github.com/wso2/labs-agentic-engineer/issues/575)) can render a log but not a
checklist.

## Users

**Developer/User**, watching any long agent turn — which, with the interview cap removed, is now
the most repeated experience in the product.

## Experience walkthrough

1. **A skill declares what it is about to write**, as a tool call the console renders — the
   convention **ADR-0012** already established for question cards. No new transport.
2. **The rail shows those artifacts as `planned`** (ghosted), then each turns *writing* and *done*
   as the run proceeds.
3. **A count appears where it is honest** — *Design · 2 of 6* — which is the answer to "how long do
   I wait".
4. **The plan grows in stages.** `design` declares the cell first; only once the cell exists does
   the component set, and so the per-component files, come into being. The list filling out in
   waves is the real shape of the work, not a glitch.
5. **A failed artifact shows as `error`** and is recovered by re-running `/design` as a delta
   pass — not a per-file retry.

## Scope

**In:** the declaration tool call and its schema; the agents-service side that lets a skill emit it;
console rendering of planned → writing → done → error into the rail; the count; staged expansion;
declaring the plan in `design` and `start`.

**Out (explicitly):**
- **The rail itself** — [#575](https://github.com/wso2/labs-agentic-engineer/issues/575) ships the
  states; this fills them.
- **Per-file regeneration.** Rejected: artifacts derive from each other and the cell fixes the
  component set everything hangs off. Recovery is a `/design` delta pass.
- **Narration wording** — the console narration skill's issue.

## Decisions

Grilled as [Waiting, and why the button is off](https://github.com/wso2/labs-agentic-engineer/issues/527).
Recorded under **The spec view's artifact rail** in [the lexicon](https://github.com/wso2/labs-agentic-engineer/blob/main/apps/console/design/lexicon.md).

- **Tool-call-as-UI, not a new event type.** ADR-0012 set the convention for question cards; a plan
  is the same shape of thing.
- **Turn-scoped, never project-scoped.** Declaring design artifacts during a *requirements* turn
  would recreate the `Being derived…` defect
  [#523](https://github.com/wso2/labs-agentic-engineer/issues/523) removed — *planned* means
  **about to be written now**.
- **The plan may grow mid-run**; it is not a contract the agent is held to for the whole turn.
- **Emergent-only progress plus elapsed time** was the cheaper alternative and was rejected: it can
  never answer "how much longer", which is the question being asked.

## Contract changes

**Yes — needs a BE handshake.** A tool the design-time agent can call to declare a plan, and its
shape (artifact identity, ordering, whether an entry may be added later). It is agent-facing rather
than REST, so confirm whether it belongs in the OpenAPI contract at all or only in the
agents-service tool schema plus `@aep/agent-stream`.

## States to design

- A declared-but-not-started plan (all ghosts).
- Mid-run with the plan growing — new entries appearing below completed ones.
- A turn that dies partway: some done, one errored, the rest still ghosts.
- A skill that declares nothing (no plan) — the rail must degrade to today's behavior.

## Open questions

Whether `start` benefits from this at all. Its plan is the PRD's sections, and it writes one file —
the count may be noise there. Worth deciding during the build rather than up front.

hevayo:
## What this does

The skills know their own plans — `/design` walks a known artifact set — but that knowledge never reached the wire. The spec rail could log what had *happened* and never say how much was *left*: a 90-second design run had no denominator anywhere on screen.

**`declare_plan`** joins the wire contract: the agent says what it is about to write, and the rail renders it.

Closes #576.

## The shape

**A new class of tool call (ADR-0025).** ADR-0012 established tool-call-as-UI for question cards, where the turn *ends* at the call (`hasToolCall` stop) because a human is being asked something. A plan declaration is the same transport and the opposite intent: `execute` resolves immediately, no stop condition, the agent keeps working. Everything else in ADR-0012 carries over — the existing `tool-call` frame, no new SSE event kinds, `@aep/agent-stream` owning the wire types so producer and renderer cannot drift, and rehydrate-from-replayed-history.

**A plan entry is a bare ordered path.** No display labels ride the wire: naming a document is the console's job (the same rule #580 set for narration), and the path is what the mutation stream already carries, so reconciliation is exact string match.

**Union semantics, no removal.** The plan *grows* — the design agent writes the cell first, and only the cell fixes the component set — so later calls add the per-component files. Repeated declarations accumulate: first-seen order kept, restated paths ignored. This is the one rule robust to an agent restating its whole plan; under append-only that would duplicate rows, under replace it would silently *shrink* the count.

**Every status is derived, never self-reported.** `writing` when the file's input stream opens and resolves its path, `done` when it closes, `error` only for the entry being written when the turn died. The agent declares intent; the stream decides what actually happened.

**A clean turn's plan dissolves; a dead turn's wreckage persists.** When everything lands, the files *are* the record and the count goes away. When a turn dies, the residue — done ticks, the one error, the remaining ghosts — stays until the next declaring turn replaces it, because that residue is the case for re-running. It surfaces through the section's existing attention chip: *The design run didn't finish* → **Update the design**. Recovery is the `/design` delta pass, never a per-file retry.

**No progress narration.** Stage prose in the chat was in scope, was built, was tested on the live stack, and is **withdrawn**: watching the artifacts appear as they are written already tells the user what is happening, so a line announcing each step restates the rail in words. The chat speaks when it has something the rail and editor cannot carry — a decision, a question, a failure. Recorded in the lexicon under the three surfaces, and on the issue.

**A question mid-flight pauses the plan, it does not end it.** A design run that stops to ask something continues in the turn the answer opens — same work, new turn id — so the plan is *paused* and adopted by that turn rather than discarded. Paused is neither done nor failed, so reloading while a question is open never reads as failure. The continuation is decided from **state** (is a question outstanding?), never from the message text: ADR-0012 makes free text in the composer an equally valid answer, so a marker-matching rule silently failed for anyone who typed instead of clicking.

**The editor follows the write (ADR-0026).** Each artifact is selected as its write starts, in whatever renderer it already has — the cell as the Architecture diagram, criteria as the structured view. The first manual selection is a declaration of reading intent and ends following for the rest of the turn; the rail's pulse stays the one-click way back. This **supersedes** the cell's burst navigation, which yanked back even over a manual selection — a six-file run would have been six yanks.

## Proof of execution

Driven in the running console (`VITE_API_MODE=mock`) via the `/design` mock stream, which exercises both publish paths (streamed input *and* the complete tool-call), the union dedupe, and a `fail` variant that dies mid-write.

| # | shows |
|---|---|
| `576-1-plan-declared.png` | Wave one declared (*Planned 3 documents*), the cell created, then wave two. **The dedupe is visible in the UI**: wave two declares 4 paths and the chat reads *Planned 3 more documents* — the restated `design.md` was ignored. Rail at *2 of 5*, **Security** still a dimmed ghost, and the pane says the next document is on its way. |
| `576-2-mid-run.png` | Further in: *5 of 5*, Security now a real row, the acceptance criteria still being written. |
| `576-3-done.png` | **The plan dissolved** — Design back to a plain green tick with no count, no ghosts left. The editor followed the last write to the acceptance criteria, rendered in its own structured view rather than as text. |
| `576-4-wreckage.png` | `/design fail` died mid-write: Design amber at *4 of 5* with a counted warning chip; the entry the turn died on carries the error, and the pane correctly stops claiming a document is coming now the turn has ended. |
| `576-5-wreckage-dialog.png` | The chip's dialog: *The design run didn't finish* → **Update the design**. |

## Review round

`/code-review` on the branch raised 8 findings; all 8 are fixed in the second commit. The one worth calling out:

**A clean turn rehydrated as permanent wreckage.** The rehydrate read the last assistant message carrying a `declare_plan` and looked for the file writes *beside it* — but the AI SDK appends one assistant message per **step** (`run-turn.ts` pushes the whole of `result.responseMessages`), so a turn's declaration and the writes it predicted are never in the same message. Every reload after a *successful* design run would have shown *The design run didn't finish* over files that exist. Now accumulated per turn across all its steps, bounded by user messages. The unit test that passed before only did so because it put both calls in one message — a good reminder that a test written from the same wrong mental model as the code proves nothing.

The rest: `removeFile` drove the plan lifecycle (following a document about to vanish, then ticking it green); the wreckage reason claimed a "design run" on Requirements and offered a button that would re-derive the design *from* the requirements the dead turn never finished; follow-the-write hit a dead-end "Select a file to view its content." in the window before a body reaches the room; an errored row rendered enabled over no file; the Security row read only one of its two paths; and ghosts were appended rather than merged in path order, so rows hopped when a write landed.

Also corrected a code comment claiming a drift guard that does not exist yet, and recorded the producer's absence in ADR-0025 rather than leaving the ADR implying a registered tool.

## Tests

- `planStore.test.ts` (13) — union + dedupe, the derived lifecycle, clean dissolve vs. wreckage, a failed turn whose plan all landed leaving nothing, cross-turn replacement, parse tolerance, and rehydrate (rebuild, dissolve, never clobber a live fold).
- `runTurn.test.ts` (+3) — the belt-and-braces double publish collapsing to one row, statuses derived from real file frames, committed turn dissolving the plan.
- `railSections.test.ts` (+5) — the per-section count, the writing entry outranking the flow-token guess for the active pulse, wreckage as an attention reason (including on a section with nothing committed).
- `SpecFileList.test.tsx` (+2) — disabled ghost rows, the header count.
- `SpecView.test.tsx` — the three follow-the-write rules, replacing the two tests for the superseded cell yank-back.

Plus regression tests for every shape found by testing on the live stack — a design run spanning a question and answer, a free-text answer, several question rounds, and a genuinely new instruction still starting new work — and six from the review round — the real multi-step transcript shape, the two-wave union across steps, turn-boundary isolation, `removeFile` settling nothing, the disabled errored row, and the waiting state appearing (and correctly stopping once the turn ends).

Full suite: **125 files, 1498 tests passing** (the count rose with the merge from `main`). `lint`, `tsc`, and `make deadcode-ts-check` clean. Every screenshot above was re-captured *after* the fixes.

## Contract

`declare_plan` in `@aep/agent-stream` only — **no `aep-api` change**. The plan is turn-scoped ephemera on the existing SSE stream; nothing RESTful ever queries it, and a reload rehydrates from the same conversation history the question cards replay from.

## Both halves are here

There is no backend handshake left to open — the producer landed on this branch:

- **`declare_plan` registered** in the agents service beside `draftExternalResource` (already fire-and-forget), with the Zod schema and the `Equal<>` drift guard every other input carries, so the wire type and the validator can no longer split. Deliberately **not** paired with a `hasToolCall` stop condition: the agent declares and keeps working.
- **The `design` skill declares before it writes**, and again when the plan grows.

A skill that declares nothing still works — the rail simply degrades to its pre-#576 behaviour, which is also the permanent path for `start`.

## Docs

- **ADR-0025** — fire-and-forget UI tool calls (amends ADR-0012).
- **ADR-0026** — the editor follows the write, and yields to the reader (supersedes the #575-era cell navigation).
- `lexicon.md` — the artifact-state table filled in with what was actually built.
- `PRD.md` — feature inventory entry.

## Merged `main`

`main` since this branch opened collapsed the security design onto
`security.json` alone (#665) and gave the project overview an
`?view=architecture` link into the workspace (#662). Both are kept:
`followSelection` and the Security rail row now speak for the ONE security
path (the "more advanced of the two files" rule went with the split it existed
for), and the `?view=architecture` one-shot survives beside follow-the-write
while the cell's burst navigation stays retired. `SecurityPanel` takes main's
read-only props, and the mock design run declares and delivers a real
`security.json` so the mock-mode proof still lights the Security row.

The two ADRs here are renumbered **0025** and **0026** — `main` landed
ADR-0022 (the overview is a track of links) in the same directory, and
0023/0024 are claimed by another branch in flight.

Full suite after the merge: **130 files, 1611 tests passing**; `lint`, `tsc`
and `make deadcode-ts-check` clean.


hevayo:
## CodeRabbit review — all 7 addressed

Checked each against current code rather than the code as reviewed (several commits landed after the review). **All seven were still valid, and all seven are fixed** in `788f672c`. Three were substantive:

**1. A write was ticked before the bundle ruled on it.** The plan settled at `tool-input-end` — the body being complete — so a write the write-gates went on to REJECT showed a green tick on the rail. `chatStore.ts` carries a comment explaining precisely why the chat card must not do this (*"guessing true would show a success tick on a write the gates may still reject"*) and the plan did it anyway. Entries now stay `writing` until the tool-result; a rejected write settles as `error`.

Rehydrate had the same fault from the other side, which the review also caught: it settled from the tool-**call**, and a failed write leaves one of those too — so a refused write came back green after a reload. It now reads the result's `ok`.

**2. Wreckage was cleared by any later turn.** Decision 6 says the residue stands until the next *declaring* turn replaces it, but `forTurn()` started fresh on any new turn id — so a single unrelated write silently dropped the alarm. Wreckage is now carried across an ordinary turn, and only a declaration replaces it. Its writes still count: writing out the last outstanding entry clears the wreck properly, which is the honest way for it to go.

**3. A document link inside the editor bypassed follow-the-write.** The in-editor reference links called `setSelection` directly rather than `selectManually`, so a reader who clicked a reference mid-turn kept being pulled away by the next write. It is manual navigation and now ends following like any other pick (ADR-0023).

The other four:

- **Security row** now reports the more *advanced* of its two files (writing beats merely-planned), instead of keying off `security.md` alone.
- **Ghost-row test** identifies the ghost and asserts exactly one row is disabled — it previously accepted *any* disabled button in the nav, so it passed for the wrong reason.
- **Wire comment** now distinguishes the outcomes: a clean finish dissolves the plan whole, a failure keeps the remainder standing, and a question mid-flight ends neither.

Nothing was declined. Six new tests cover the corrected semantics — verdict-settles-the-entry, a rejected write as an error, wreckage surviving an ordinary turn and being cleared by completion, replacement by a new declaration, and the rejected-write rehydrate.

Console **1504** tests, agents **321**, lint + knip clean.

kaje94 (dismissed):

kaje94:
we have the screenshots in the file changes.
Shall we remove them?

kaje94 (approved):

PR #644, opened 2026-08-28
feat(skaffold): full local dev path via skaffold dev

senithkay:
## Summary

- **Self-contained local dev**: `skaffold dev` now builds all 7 service images from monorepo source, deploys the platform Helm chart, and runs Thunder OAuth registration — no manual steps after `aectl platform install`
- **All images built locally**: aep-api, agents, collab, mcp-server, console, remote-worker, thunder-app-operator — imported into k3d via `local.push=false`
- **Addons bundled**: thunder-app-operator and cnpg are now Helm releases in the skaffold path (previously aectl-only optional addons)
- **Secrets handled safely**: `skaffold/secrets.yaml` is gitignored (contains real Thunder admin creds per cluster); `skaffold/secrets.yaml.example` is the committed template with placeholders
- **Thunder 0.34 compatibility**: fixed nested sqlite config schema (`sqlite.path`/`options`) and added `readOnlyRootFilesystem: false`

## Key files

| File | What changed |
|---|---|
| `skaffold.yaml` | Full build artifacts + Helm releases + before/after hooks |
| `skaffold/generate-and-apply.sh` | Before hook — applies namespaces/secrets, auto-generates RSA signing key |
| `skaffold/thunder-setup.sh` | After hook — registers all AEP OAuth clients in Thunder via Bearer auth |
| `skaffold/secrets.yaml.example` | Committed template; copy to `secrets.yaml` and fill in Thunder admin creds |
| `skaffold/addon-namespaces.yaml` | Namespace for thunder-app-operator-system |
| `deployments/single-cluster/values-thunder.yaml` | Thunder 0.34 config fixes |
| `skaffold/configs.yaml` | Deleted — values moved to `setValues` in skaffold.yaml |

## Notable fixes

- `codingAgentRunner.image` uses `REPO:TAG` template variable concatenation — `IMAGE_*` full-form resolves to empty in Skaffold v4beta11
- Thunder OAuth setup uses `POST /oauth2/token` Bearer auth (not basic auth) and handles APP-1020 name conflicts idempotently
- Helm `before` hook uses `dir: "."` to ensure correct CWD for `kubectl apply -f skaffold/...` paths

## Test plan

- [ ] Copy `skaffold/secrets.yaml.example` → `skaffold/secrets.yaml`, fill in Thunder admin creds
- [ ] Run `skaffold dev --kube-context k3d-openchoreo`
- [ ] Verify all pods reach Running state in `wso2-aep`
- [ ] Verify Thunder OAuth clients registered: port-forward Thunder and `GET /applications`
- [ ] Log in to AE console at `http://console.openchoreo.localhost:8080`


skaffold.yaml
```diff
@@ -1,101 +1,74 @@
...
+      docker:
+        dockerfile: Dockerfile
+  local:
+    push: false
+
+manifests:
+  rawYaml:
+    - skaffold/namespace.yaml
+    - skaffold/addon-namespaces.yaml
+    - skaffold/secrets.yaml
```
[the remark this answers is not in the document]
senithkay: dev is expeccted to copy the example file and create his own yaml

skaffold.yaml
```diff
@@ -1,101 +1,76 @@
...
+      docker:
+        dockerfile: Dockerfile
+        cliFlags:
+          - "--build-context=skills=./skills"
+    - image: ghcr.io/wso2/aep/thunder-app-operator
+      context: deployments/single-cluster/resource-types/thunder-app/operator
+      docker:
+        dockerfile: Dockerfile
+  local:
+    push: false
```
kaje94, line 48: `useBuildkit: true` was dropped here, and I think it breaks the builds for `agents`, `aep-mcp-server` and `console`.

Without it, Skaffold builds any artifact that has no `cliFlags` through the Docker Engine API, which uses the old builder. Those three Dockerfiles all use `RUN --mount`, and that only works with BuildKit.

Ran this on the branch with Docker 29.2.1 (its CLI defaults to BuildKit, so it isn't an old-Docker problem):

```
$ skaffold build -b ghcr.io/wso2/aep/aep-mcp-server
Sending build context to Docker daemon  439.8kB
Step 5/17 : RUN --mount=type=cache,id=pnpm-store,target=/pnpm/store ...
docker build failure: the --mount option requires BuildKit.
```

`aep-api` and `remote-worker` are fine, but only because `cliFlags` makes Skaffold use the docker CLI instead.

Could we put `useBuildkit: true` back? The comment that was deleted with it described this exact problem.

skaffold.yaml
```diff
@@ -1,101 +1,76 @@
...
     - image: ghcr.io/wso2/aep/console
       context: .
       docker:
         dockerfile: apps/console/Dockerfile
+    - image: ghcr.io/wso2/aep/remote-worker
+      context: runners/remote-worker
+      docker:
+        dockerfile: Dockerfile
+        cliFlags:
+          - "--build-context=skills=./skills"
```
kaje94, line 35: `remote-worker` needs two build contexts, not one.

Its Dockerfile uses `bal-library-tool` (lines 54-58 and 67) as well as `skills` (line 205). `deployments/scripts/build-runner.sh` passes both.

With only `skills` passed, Docker treats `bal-library-tool` as an image name and fails:

```
ERROR: failed to solve: bal-library-tool: failed to resolve source metadata for
docker.io/library/bal-library-tool:latest: pull access denied, repository does not exist
```

So it needs one more flag:

```yaml
- "--build-context=bal-library-tool=./packages/bal-library-tool"
```

skaffold.yaml
```diff
@@ -1,101 +1,76 @@
...
+      docker:
+        dockerfile: Dockerfile
+  local:
+    push: false
+
+manifests:
+  rawYaml:
+    - skaffold/namespace.yaml
+    - skaffold/addon-namespaces.yaml
+    - skaffold/secrets.yaml
```
kaje94, line 55: This file is gitignored, so a fresh clone won't have it.

Skaffold validates the manifest list before it runs the `before` hook, so `generate-and-apply.sh` never gets the chance to create it. On a clean checkout:

```
$ skaffold render
invalid skaffold config: ... in module "aep" on line 45 column 5: Manifest file
"skaffold/secrets.yaml" referenced in skaffold config could not be found.
```

It can't be generated automatically since it needs real Thunder admin credentials. Two options that would work:

- have `generate-and-apply.sh` copy the example and stop with a clear message if the placeholders are still in it, or
- drop it from this list, since the hook already applies it with `kubectl`.

skaffold/thunder-setup.sh
```diff
@@ -0,0 +1,304 @@
...
+}))" "$name" "$desc" "$OU_ID" "$cid" "$secret")
+  thunder_upsert_app "$cid" "$payload"
+}
+
+# ── Confidential clients ─────────────────────────────────────────────────────
+echo "Registering AEP OAuth clients in Thunder..."
+thunder_confidential "Workload Publisher"                  "OC Workload Publisher Client"              "openchoreo-workload-publisher-client"       "${OC_WORKLOAD_PUBLISHER_SECRET}"
+thunder_confidential "OpenChoreo Observer Resource Reader" "BFF token for OC Observer service"         "openchoreo-observer-resource-reader-client" "${OC_OBSERVER_READER_SECRET}"
+thunder_confidential "AEP API Service"                     "AEP API service-to-service client"         "aep-api-client"                             "${AEP_API_CLIENT_SECRET}"
+thunder_confidential "AEP BFF to git-service"              "BFF outbound JWT, audience: git-service"   "bff-git-service"                            "${BFF_TO_GIT_SERVICE_SECRET}"
```
kaje94, line 209: These three client IDs don't match the ones the Thunder bootstrap registers:

| here | `values-thunder.yaml` / `setup-local.sh` |
|---|---|
| `bff-git-service` | `aep-bff-to-git-service` |
| `bff-remote-worker` | `aep-bff-to-remote-worker` |
| `local-dev-seeder` | `aep-local-dev-seeder` |

The short names come from aectl (`tools/aectl/cmd/thunder_cli.go`), so the mismatch isn't new. But aectl sets the app *name* to the same value as the client ID, so it just ends up creating extra apps. Here the app names are the long ones from `setup-local.sh`, so they clash with the bootstrap apps. That makes the `APP-1020` branch below (lines 150-157) delete the bootstrap app and recreate it — on every fresh cluster, not just an old one.

Two knock-on effects:

- `seed-dev.sh` looks for `aep-local-dev-seeder`, so seeding stops working. Same for `setup-aep.sh`, `scripts/project-rebuild.sh` and `scripts/project-revalidate.sh`.
- Running `make setup-local` afterwards fails with `APP-1020 but could not find existing app` and exits 1.

Smallest fix is to use `aep-local-dev-seeder` here. The secret in `secrets.yaml.example` already matches what those scripts expect, so only the ID is off.

For the other two, I couldn't find anything that uses them at all, in either spelling. The BFF signs its own JWT to reach git-service (`IssueServiceToken`, `iss=aep-bff`, `aud=git-service`) rather than using a Thunder client. `SERVICE_AUTH_GIT_CLIENT_ID` in `docker-compose.yml` isn't read anywhere, and `BFF_TO_*_SECRET` never reaches a pod. If that's right they could be deleted everywhere instead of renamed, which is probably its own cleanup issue. Worth noting the same three IDs came up in the PR #473 review as well.

skaffold/generate-and-apply.sh
```diff
@@ -0,0 +1,47 @@
...
+if kubectl get secret aep-task-signing-key -n "$NAMESPACE" >/dev/null 2>&1; then
+  echo "aep-task-signing-key already exists — skipping"
+else
+  echo "Generating aep-task-signing-key..."
+  RSA_KEY=$(openssl genrsa 2048 2>/dev/null)
+  kubectl create secret generic aep-task-signing-key \
+    --namespace="$NAMESPACE" \
+    --from-literal="task-signing.pem=$RSA_KEY"
+  echo "aep-task-signing-key created"
+fi
```
kaje94, line 47: `setup-local.sh` also stamps Helm ownership onto a few resources that `setup-aep.sh` creates before the chart is installed, and that step isn't here.

Two of them clash for certain:

- `ClusterTrait/api-configuration` — the chart defines it in `templates/openchoreo-org-types/clustertrait-api-configuration.yaml`, and `setup-aep.sh` applies `deployments/manifests/api-platform/api-configuration-trait.yaml`
- `ClusterAuthzRoleBinding/aep-api-client-binding` — chart `templates/aep-api/authz.yaml`

Without the `meta.helm.sh/*` annotations, `helm install` stops with `invalid ownership metadata`.

Was leaving this out deliberate? Asking because the description says there are no manual steps left, but the `setup-local` target is still in the Makefile.

skaffold.yaml
```diff
@@ -107,10 +82,44 @@ deploy:
...
+          # platform.workspaces.access_mode
+          workspaces.accessMode: "ReadWriteOnce"
+      - name: thunder-app-operator
+        chartPath: deployments/single-cluster/resource-types/thunder-app/operator/helm
+        namespace: thunder-app-operator-system
+        createNamespace: true
+        setValueTemplates:
+          image.repository: "{{.IMAGE_REPO_ghcr_io_wso2_aep_thunder_app_operator}}"
+          image.tag: "{{.IMAGE_TAG_ghcr_io_wso2_aep_thunder_app_operator}}"
+      - name: cnpg
```
kaje94, line 125: `setup-prerequisites.sh` already installs a release called `cnpg` into `cnpg-system`, at the same version it reads from `env.sh`.

So with this, Skaffold will upgrade a release it didn't create, and a `skaffold delete` would remove something the rest of the cluster depends on. The version is also written down in two places now, so it's two places to bump.

skaffold.yaml
```diff
@@ -1,101 +1,76 @@
...
         valuesFiles:
-          - deployments/helm-charts/platform/values.yaml
-          - deployments/helm-charts/platform/values.local.yaml
-          # Per-developer overrides (git-ignored), layered last so it wins. Copy
-          # values.local.dev.yaml.example to values.local.dev.yaml and edit — e.g.
-          # webhook.deliveryURL for local GitHub webhook delivery. The
-          # dev-cluster/deploy-local Make targets create an empty stub if it is
-          # missing, so skaffold never fails on a fresh checkout.
-          - deployments/helm-charts/platform/values.local.dev.yaml
+          - skaffold/helm-values.yaml
```
kaje94, line 81: A few leftovers from moving the overrides to `skaffold/helm-values.yaml`:

- The Makefile still creates a `values.local.dev.yaml` stub that this config no longer reads, and it still runs on both `dev-cluster` and `deploy-local`.
- `values.local.yaml` becomes unused — this was its last reference.
- `deployments/README.md` still describes `values.local.dev.yaml.example` as the Skaffold override template.
- The `portForward` for aep-api:9090 is gone, but the root `README.md`, `deployments/README.md` and the Makefile all still mention it.

skaffold/thunder-setup.sh
```diff
@@ -0,0 +1,304 @@
...
+THUNDER_NS="${THUNDER_NS:-thunder}"
+THUNDER_DEPLOYMENT="${THUNDER_DEPLOYMENT:-thunder-deployment}"
+THUNDER_CONFIG_MAP="${THUNDER_CONFIG_MAP:-thunder-config-map}"
+CONSOLE_URL="${CONSOLE_URL:-http://console.openchoreo.localhost:8080}"
+
+# PyYAML is required for the Thunder CORS config patch.
+# Install it if missing rather than failing mid-hook after OAuth registration.
+if ! python3 -c "import yaml" 2>/dev/null; then
+  echo "PyYAML not found — installing..."
+  pip3 install --quiet pyyaml || { echo "ERROR: could not install PyYAML. Run: pip3 install pyyaml"; exit 1; }
```
kaje94: Small one: this installs a package into the developer's Python without asking.

On newer Homebrew and Debian Pythons it also fails outright with `error: externally-managed-environment`, so the hook would stop here.

Could the CORS patch read the ConfigMap with `kubectl -o jsonpath` and `yq` instead? Or just print a message asking the user to install PyYAML rather than doing it for them.

kaje94 (commented):
Found 3 blockers. The rest are smaller points.

Blockers:
1. Removing `useBuildkit: true` breaks 3 of the 7 image builds.
2. `remote-worker` is missing its second build context.
3. A fresh clone fails straight away because `skaffold/secrets.yaml` is gitignored.

Two notes on the PR itself: the description mentions `deployments/single-cluster/values-thunder.yaml` and a deleted `skaffold/configs.yaml`, but neither is in the diff. And the branch is 121 commits behind `main`, so a rebase before running the test plan is probably worth it.

skaffold/thunder-setup.sh
senithkay: Addressed in a5363a5f53388abb8add26547923ec21d1629667

skaffold.yaml
senithkay: Scripts will be refactored in a different PR

skaffold/generate-and-apply.sh
senithkay: These scripts will be refactored in https://github.com/wso2/labs-agentic-engineer/pull/676

skaffold.yaml
senithkay: Its a dev responsibility to make there own secrets.yaml file

skaffold.yaml
senithkay: a5363a5f53388abb8add26547923ec21d1629667

skaffold.yaml
senithkay: Addressed in 16dbc2f9c55a1804dfd68f2b3590210ca6e75ced

skaffold/thunder-setup.sh
```diff
@@ -0,0 +1,304 @@
...
+print(json.dumps({
+  'name': name, 'description': desc, 'ouId': ou_id,
+  'inboundAuthConfig': [{'type': 'oauth2', 'config': {
+    'clientId': cid, 'clientSecret': secret,
+    'grantTypes': ['client_credentials'],
+    'tokenEndpointAuthMethod': 'client_secret_post',
+    'pkceRequired': False, 'publicClient': False,
+    'token': {'accessToken': {'validityPeriod': 3600}}
+  }}]
+}))" "$name" "$desc" "$OU_ID" "$cid" "$secret")
```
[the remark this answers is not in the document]
senithkay: intentionally kept this way

skaffold.yaml
senithkay: a5363a5f53388abb8add26547923ec21d1629667

skaffold/thunder-setup.sh
senithkay: Leaving this as-is — PR #676 makes the mismatch moot. It deletes `seed-dev.sh`, `setup-aep.sh`, and the other scripts that look for `aep-local-dev-seeder`, and removes the `aep-local-dev-seeder` entry from `values-thunder.yaml` entirely. Once #676 lands there is nothing that conflicts with the short IDs registered here and nothing that depends on the long name, so the APP-1020 delete-and-recreate issue disappears too.

skaffold/thunder-setup.sh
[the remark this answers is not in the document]
senithkay: aectl is ment to be prod friendly and the other path is for the local dev. we should not couple them

PR #730, opened 2026-09-07
Make Oxygen UI the organization's design system, keeping Astryx available

axewilledge:
## Problem

The library shipped one design system, `astryx-design-system`, and `organization` named it. This org builds its own UI toolkit — [WSO2 Oxygen UI](https://github.com/wso2/oxygen-ui) (`@wso2/oxygen-ui`) — and every `web-application` the platform generates should be built with it.

## Changes

`skills/AGENTS.md` already documents what a swap costs: two edits, both in places an org may edit. This is those two edits, plus the skill the first one requires.

- **New `skills/oxygen-ui-design-system/`** — declares the four things "Swapping the UI design system" requires of the slot: `kind: org` + `audience: [coding]`, a `## Verify` command, the platform facts that override the vendor's own docs (no `base`, no extra `index.html` tags, never run `npx @wso2/oxygen-ui init`, never install the peer set its README lists), and its ownership boundary against `react-webapp` (Oxygen owns UI under `src/`; the `openapi-fetch` + `src/generated/` data layer stays `react-webapp`'s).
- **`organization`** — the **UI design system** section now names `oxygen-ui-design-system`. That one line is the switch: `architecture` reads the name from the Organization defaults block and pins it on every `web-application`, so no platform skill changed.
- **`astryx-design-system`** — kept, one line changed: its description no longer claims to be *this organization's* design system, because `organization` no longer names it.
- **`skills/AGENTS.md`** — records that the library now ships two design systems and which is the default, and widens the vendor-name guard.

## What "keep Astryx disabled" does and does not do

Worth being precise, because the library cannot do the whole of it.

Astryx is no longer **pinned**, so no build is steered by it — that half is done and it is the half that matters for output. But it is still seeded **enabled**: per `ADR-0015` an absent manifest entry means enabled, and `reconcileEmbedded` only carries a `Disabled` flag *forward* from an org's own repo, never originates one. So Astryx still lands in each project mirror and a coding agent could still load it by name.

**Turning it off is one toggle per org at Settings → Skills**, which is where ADR-0015 deliberately put availability. Nothing in this PR can pre-set it, and I did not try to fake it by narrowing `audience` — audience is visibility, not availability, and narrowing it would break any component that legitimately pinned Astryx. `skills/AGENTS.md` now states this outright rather than implying the skill is inert.

## Verify is a script, not a vendor CLI

`react-webapp`'s verify sequence reserves one slot for the design system, and Astryx fills it with `astryx doctor`. Oxygen ships no equivalent, so the skill ships `scripts/verify.mjs` — precedent: `mock-verification/scripts/walk.sh`, `aep-validation/scripts/generate-report.mjs`. It catches what `tsc` and `vite build` cannot, because these type-check and build perfectly clean and then render an unthemed or broken page in the cluster:

- a `@mui/*`, `@emotion/*` or `lucide-react` installed directly beside the copies Oxygen and its icons package already bring (two Emotion caches — the theme reaches half the tree)
- an import that bypasses `@wso2/oxygen-ui`, in `src/` **or in the build config**
- a root not wrapped in `OxygenUIThemeProvider`
- a React that is not the exact version Oxygen's peer dependency pins (two Reacts; hooks throw at runtime)

Each failure line names its fix. `scripts/verify.test.mjs` pins one fault per check (12 tests) and rides `make test` via the `skills/**/*.test.mjs` lane.

## The skill points at the package's docs instead of copying them

First draft vendored the package's `components.md` / `patterns.md` / `theming.md` (2,972 lines). Review caught that against `skills/AGENTS.md`, "Who owns what": *"A tool documents itself; a skill points at it … the two ship on different clocks … so a copy here goes stale with nothing to catch it."* It was already stale — the copy carried symbols 0.13.1 does not export.

Those three are gone. `@wso2/oxygen-ui` ships its own agent docs inside the package (`.claude/*.md`, guaranteed by its `files` field), so the skill points there and at the `.d.ts` for anything the prose does not settle. Only `references/app-structure.md` remains, because it is the platform-shaped scaffold rather than a vendor copy. This also matches `astryx-design-system`, which ships no references at all.

The skill does carry a short **Known errors in the package's docs** table, for three snippets that do not compile against the code they ship with — each confirmed with `tsc`, not asserted:

| The docs show | It actually is |
|---|---|
| `<Footer companyName="WSO2 LLC" />` | `Footer` takes children |
| `DashboardIcon` | not an export — `LayoutDashboard` |
| `GoogleIcon` | not an export — brand icons carry no `Icon` suffix |

## Verification

Proved against a **real app** scaffolded with the skill's own Setup commands and built:

```
verify.mjs        ok (8 checks)
npx tsc --noEmit  exit 0
npm run build     ✓ built
```

Negative cases behave: a stray `@mui/material` import in `src/`, and an `@emotion/*` pulled into `vite.config.ts`, each flip exactly one check and fail the step.

Every factual claim about icon names was settled by type-checking the names themselves rather than by reading docs — which is how the draft's inherited *"there is no `EditIcon`"* rule was caught and removed. lucide resolves bare **and** `Icon`-suffixed (`Pencil`, `PencilIcon`); only the brand icons have no alias.

Suites: `skills/**/*.test.mjs` 34/34, remote-worker 449/449, agents 347/347, playground 98/98.

`services/aep-api ./internal/spec` has one failure, `TestLoadEmbeddedLibrary`, that **fails identically on `upstream/main`**: `skills/agent-browser/SKILL.md` has an unquoted `description:` containing a colon, so `loadLibrary` cannot parse it and drops the skill from every org. Pre-existing and out of scope here — filed separately.

Vendor-name guard passes, with one exception now documented in `skills/AGENTS.md`: `wireframes` names Oxygen for the **wireframe compiler's** drawing palette, which is not the app's UI toolkit and does not follow the org's design system.

### DIFF
--- a/skills/AGENTS.md
+++ b/skills/AGENTS.md
@@ -139,13 +139,13 @@ changing a web-app skill:
 grep -rniE 'astryx|@astryxdesign|\boxygen\b|@wso2/oxygen' skills/ --include='*.md'
 ```
 
-Only `skills/organization/SKILL.md`, `skills/oxygen-ui-design-system/**`,
-`skills/astryx-design-system/**` and this file should match. A hit anywhere
-else — especially in `architecture`, which is `kind: platform` and read-only in
-the console — means an org can no longer swap its design system without a
-platform change.
+Four paths may match, plus the one exception below: `skills/organization/SKILL.md`,
+`skills/oxygen-ui-design-system/**`, `skills/astryx-design-system/**` and this
+file. A hit anywhere else — especially in `architecture`, which is
+`kind: platform` and read-only in the console — means an org can no longer swap
+its design system without a platform change.
 
-One documented exception: `skills/wireframes/SKILL.md` says the wireframe
+The exception: `skills/wireframes/SKILL.md` says the wireframe
 compiler renders with an "Oxygen UI palette" and applies "the Oxygen theme".
 That is the **compiler's** drawing style for a `.excalidraw` picture, not the
 app's UI toolkit, and it does not follow the org's design system — swapping
--- a/skills/AGENTS.md
+++ b/skills/AGENTS.md
@@ -43,8 +43,8 @@ An absent kind means `org`, which is a real decision, not a default to lean on:
   `aep-validation`, `mock-verification`) and the browser CLIs they drive
   (`playwright-cli`, `agent-browser`).
 - **`org`** — the org-visible stack skills (`go`, `ballerina`, `react-webapp`,
-  `astryx-design-system`, `api-management`, `thunder-authentication`). Editable
-  and deletable by an org.
+  `oxygen-ui-design-system`, `astryx-design-system`, `api-management`,
+  `thunder-authentication`). Editable and deletable by an org.
 
 Kind decides console visibility and who may edit a skill, and nothing else. It is
 `audience` that decides who may *read* one, and the two are independent — a skill
@@ -84,9 +84,13 @@ one was `go`'s, and it was invisible for as long as `go` was preloaded regardles
 A web app's UI toolkit is an **organization** decision, and nothing in this
 library hardcodes one. Two edits change it, both in places an org may edit:
 
-1. Add the new design-system skill — author `skills/<name>/SKILL.md`, or import
-   it through Settings → Skills. Delete `astryx-design-system` if the org does
-   not want it available.
+1. Make sure a skill for the design system exists — author
+   `skills/<name>/SKILL.md`, or import it through Settings → Skills. The
+   library ships two: `oxygen-ui-design-system` (WSO2 Oxygen UI, the default
+   `organization` names) and `astryx-design-system` (kept available, named by
+   nothing). Delete the one an org does not want offered at all, or switch it
+   off in Settings → Skills — availability is the org's manifest flag
+   (ADR-0015), not something a library file can set.
 2. Point the **UI design system** section of `organization` at its name.
 
 `architecture` reads the name out of that section rather than holding one of its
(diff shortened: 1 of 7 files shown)


axewilledge:
Addressed the review in f9abcc5e. Five findings were valid and are fixed, each pinned by a test (suite 12 → 18):

- **Import scan read prose.** `IMPORT_RE` ran over raw source, so a comment saying "never import from `@mui/material`" — advice this skill gives at length — failed a correct app. Now scans with comments blanked; a real import sharing a line with a URL is still caught.
- **Provider check accepted an import.** `includes("OxygenUIThemeProvider")` passed on the symbol being merely imported, or rendered off the root. Now requires a JSX element for it inside `createRoot().render()`.
- **`null` JSON.** `JSON.parse("null")` succeeded and the first field read then threw, killing the run before any FAIL row printed. Only a JSON object is accepted now.
- **`AppRoute` vs renderer.** Made recursive. Needed a branch: `RouteProps` is a union and an index route may not carry children — the compiler caught that when the corrected snippet was built rather than read.
- **Allowlist vs documented exception** in `skills/AGENTS.md` now agree.

Two findings are **not applicable**, left alone deliberately: `no-undef` on `process` and `no-regex-spaces` in the test file. `skills/` is deliberately not a pnpm workspace (`skills/AGENTS.md`), so `turbo run lint` never reaches it, and the precedent scripts `aep-validation/scripts/generate-report.mjs` and `mock-verification/scripts/walk.test.mjs` carry the same 11 errors. Fixing this one file would diverge it from the rest of the library for a lint run that does not happen.

Re-verified by rebuilding the proof app from the reference's own `App.tsx` verbatim: verify 8/8, `tsc --noEmit` clean, `vite build` succeeds.

xlight05 (approved):

PR #740, opened 2026-09-08
feat(design): the user chooses a provider — givens in requirements, suggestions at design, research when resolving

hevayo:
Stacked on #734 (`feat/dependency-resolution`); merge after it. The three commits at the top of this branch are this change.

## The user chooses a provider; the agent never does

Settled in an interview: consent before a service is picked, with the org's existing resources as the one exception.

- **Requirements records givens.** For each external capability the interview first lists the org's Registered External resources — a match is written as a settled decision from an org default, without a question. Otherwise it asks whether the user *already uses or must use* a service; a named answer is a settled decision ("Payments: Stripe — finance has the account"), "no preference" leaves the capability unnamed. It never proposes a provider and never tags one `*assumed*`. The requirements flows (`start`, `amend`, `settle`) now get the org catalog tool wherever they run (`catalogTurn` in `turn_runner.go`; web search stays design-only).
- **Design binds only what the PRD names** and researches it during the design turn, as before. An open capability is written as the *need*: name, purpose, and `suggestions` — services commonly used for it, from the agent's knowledge, any number, no research. No provider, no config keys (the keys follow the service; both write-gates refuse `config` before a `provider`, org copies excepted).
- **`candidates` and the `ambiguous` state are retired.** A file still carrying `candidates` reads as suggestions (Go read path and the console's parser); the fold refuses it on write with a pointer to `suggestions`. `external-ambiguous` is gone from preflight and the drawer.
- **The definition asks.** While no provider is set the view shows a **Service** card: *Which service do you want to use for this?* — a name or document URL typed + **Use this**, the suggestions as chips (*commonly used*, never *recommended*), or **Ask the agent to find one**. Every answer sends `/resolve-dependency <name> <answer>`; the trailing text reaches the skill verbatim (pinned in `turn-compose.test.ts`). The rail mark reads **Choose a service**.
- **All research lives in the resolve flow.** With an answer it settles the provider and goes for the contract; with a URL the provider is the document's `info.title`; with nothing it searches and asks ONE question when several fit, never writing options back into the file. Config keys are derived last. Providing a document through the API is choosing its system too — the provider becomes the document's title, and a document that names none is refused (`ErrDependencyNotChosen`).

Docs: ADR-0027 amended, ADR-0028, `CONTEXT.md` (`suggestions`; a new **Given** entry), the console lexicon, the spec and delivery READMEs, and the `prd-contract`, `start`, `amend`, `settle`, `architecture`, `resolve-dependency`, `design`, `organization` skills.

## Proof of execution (local stack, fresh project `small-shop-handmade`)

- **Interview** asked, beside the usual questions: *"For card payments, do you already use (or need to use) a specific payment processor?"* and the same for currency conversion, each with *Yes, we already use…* and *No preference — decide at design time*. Answered Stripe as a given, no preference for currency. PRD Product Decisions: `Payments: … through **Stripe** — the business already holds the account`; `Currency display: … no currency-conversion provider is chosen yet`.
- **Design turn** (read model after it):
  ```
  payment-provider  | unresolved needs-contract | provider: Stripe | style: rest-api | config: [STRIPE_SECRET_KEY, STRIPE_WEBHOOK_SECRET] | suggestions: []
  currency-provider | unresolved needs-input    | provider: None   | config: []      | suggestions: [Open Exchange Rates, exchangerate.host, Fixer.io]
  email-provider    | unresolved needs-input    | provider: None   | config: []      | suggestions: [SendGrid, Postmark, Amazon SES]
  ```
  The closing "Needs your input" named each open one with its options and *resolve it to pick one*.
- **Console**: rail groups with the *Choose a service* mark; `currency-provider › Definition` shows the Service card with the three chips and no Interface section. Picking **Open Exchange Rates · REST API** seeded `/resolve-dependency currency-provider Open Exchange Rates`.
- **Resolve flow**: found no published document, asked one question (write it from the docs / I'll provide one); after *write it*: `provider: Open Exchange Rates | style: rest-api | contract: openapi.yaml (marked assumed) | config: [OXR_APP_ID] | suggestions: None | provenance.sourceUrl: docs.openexchangerates.org`. The view then showed the Interface section with the acceptance box and the `openapi.yaml` link; **Accept the assumption** → `resolved`, flags `[assumed]`, the rail reads *Assumed*.
- **Build drawer** lists `payment-provider` (*No contract yet…*) and `email-provider` (*No service chosen yet — choose which one to use.*); the resolved one is gone.

Evals: `design.eval.ts` 100% (the fixture's PRD names Slack, bound at design, contract left for resolve). `requirements.eval.ts` 69% (review band) with the review sheet committed under `evals/spec-agents/eval-reviews/` — the misses are the rubric's blindness to the org's sign-in default and the sim user's own notification answer; the scenario needs no third-party service, so the new rule is exercised by the live walk above, not by this scenario.

Checks: agent-stream build/tests (179), aep-api build/vet/tests for `spec`, `platform`, `delivery/build`, `edge`, `arch` + `deadcode-check`, agents tests (348), console tsc/eslint/vitest (1782), `make license-check`, `make typecheck`. Two review passes (standards, spec) ran before the commits and their findings were applied — including dropping a read-time "provider from the document title" rule that would have made the read model disagree with the file the view and the agent read.

https://claude.ai/code/session_011VNg9Gdpd7SMv6c4VD6PLn


hevayo:
Folded into #734, which now carries the whole stack rebased on main as a single PR. Nothing here is lost: the description and proof of this part live there as its own section.

PR #750, opened 2026-09-09
Build asks in a dialog: resolve what is open, or name the version it will cut (#749)

### LINKED ISSUE #749: Build asks in a dialog: resolve what is open, or name the version it will cut
## Problem

Pressing **Build** leads to two different containers. If a dependency has no provider or no
interface on file, the console opens a right-hand drawer. Otherwise it opens a centred modal. The
user learns two surfaces for one click, and which one appears depends on state they cannot see
beforehand.

Neither says what the build will do. The drawer lists only what is blocking; the modal names the
next version and the story numbers. Nothing tells the user which components and dependencies this
version touches.

The version is always `v<N>`. The user cannot name the thing they are cutting.

## Users

**Developer/User** — owns the design gate and presses Build.

## Experience walkthrough

1. The user presses **Build**. The console commits the room's live edits and calls
   `get-build-preflight` once.
2. **Something is unresolved** — a dependency with no provider, no interface on file, or an
   org-service this project cannot see. The **Resolve dependencies** dialog opens: the names of the
   open dependencies, and one action, **Resolve**, which runs the guided flow over all of them. The
   dialog closes and the user lands in the chat. A user who prefers to work by hand opens the
   dependency's definition from the rail, as they do today.
3. **Nothing is unresolved** — the **Start build** dialog opens:
   - a version box at the top, prefilled with `v<count of versions + 1>`;
   - under it, **What changed since `v2`** — names only. A `new` chip, a `removed` chip; no chip
     means changed. A removed row states that it is out of the design and its resource stays
     provisioned, because a build deprovisions nothing;
   - fixed height, the list scrolls, the version box and the actions stay visible;
   - the primary action reads **Build v3**.
4. The user may replace the name. What they type is the tag, the milestone title, and the
   `/builds/<name>` address. Characters are restricted to what a git tag allows; a taken name shows
   an inline error and Build stays disabled.
5. **The spec tree has not changed** since the newest version: the version box is locked to the
   existing name, the list reads *no spec changes since `v2`*, and the action reads **Rebuild v2**.

## Scope

**In:**
- Two dialogs — *Resolve dependencies* and *Start build*; `BuildDependencyDrawer` is deleted.
- An editable version name that becomes the git tag.
- The change list (`new` / changed / `removed`) since the previous version.
- `get-build-preflight` extended to carry the change list and the version values.
- `spec`: tag creation date on the tag record; newest-version and unchanged-detection move off the
  number; a supplied name is used verbatim and a collision is a conflict.
- PRD **Spec versioning** amendment; `design/lexicon.md` — the *"The Build drawer lists"* entry is
  replaced.

**Out (explicitly):**
- Renaming a version after it is cut (separate work, on the version page).
- Resolving a dependency inside a dialog — the dialogs list, they never resolve.
- The full manifest of every component, external dependency and platform resource.
- The stories-in-scope line — numbers a user cannot validate.
- The per-row **Open** button in the resolve dialog.

## Decisions

1. **Both surfaces are dialogs.** The complaint was the container, not the step count. A dialog is
   for a decision that must be answered now; a drawer is for inspecting alongside your work, and
   Build cuts a tag you cannot un-cut. Rejected: keeping the drawer, because the console has one
   other drawer and seventeen dialogs.
2. **One dialog, one job.** Blocked dependencies are not rows in the build dialog: they disable
   nothing there, carry no buttons there, and the build dialog stays a single statement about the
   version. Rejected: one dialog with a `blocked` chip and a disabled Build — it mixes *you must
   act* with *here is what changed*, and only some rows would have a button.
3. **The resolve dialog has one action.** Names plus **Resolve**. Rejected: keeping a per-row
   **Open**, because a second button is a second thing to learn and the rail already reaches the
   definition.
4. **The dialogs list; they never resolve.** The user leaves, fixes the dependency on its
   definition or through the chat flow, and presses Build again. Two Build presses is accepted.
   Rejected: a workbench that rebuilds provider-picking and interface-upload inside a modal.
5. **The version name is the tag.** Free text, restricted to the characters a git tag allows. What
   the user types is the tag, the milestone title, and the address. Rejected: a name stored beside a
   fixed `v<N>`, and silent slugging of what the user typed — both break the promise that the name
   is theirs.
6. **The box is prefilled with `v<count of versions + 1>`**, incremented until the name is free.
   Rejected: highest number plus one — the count reads as "my fourth version".
7. **Order stops coming from the number.** The tag list gains the tag creation date (one more field
   in the `for-each-ref` format already used). Newest is the most recently created tag, which also
   separates two tags on one commit. Rejected: ordering by commit ancestry, and keeping a hidden
   sequence number in the tag message.
8. **A name labels a snapshot; it is not a version.** An unchanged `specs/` tree still reuses its
   version and reopens its milestone. Rejected: cutting a second tag on the same tree, which spends
   a whole planning turn to change a word.
9. **The change list is GROUPED, and a row is a name plus at most a chip** — `new`, `removed`, no
   chip means changed. Three groups, in this order, never merged, each unrendered when empty:
   **Components**; **External dependencies** (*you choose the provider and supply its keys*);
   **Platform resources** (*the build provisions these for you*). Bare names could not say what a
   name IS — `ceramics-db` and `currency-service` read identically while one is a database the
   build stands up and the other needs a provider and credentials — which is also why the two
   dependency kinds are never one group. Rejected: names with no grouping (what shipped first, and
   what the first real dialog disproved); one **Dependencies** group folding both kinds, which
   hides the only difference that changes what the user must do next; *affected artifacts* as a
   title (in the lexicon an artifact is a document in the spec rail); marking only `new` (a removed
   row would then read as changed).
9b. **The requirements are not a row.** Every row names something that EXISTS once the version is
   built, and the requirements are the input to that — they also move on nearly every version, so
   the row carried no signal while sitting among component names as though it were one. The
   `requirements` kind is gone from the contract rather than left unrendered.
10. **Changed rows are worth the diff.** Most versions change content, not membership, so a list of
    additions and removals alone would be blank for the common version. The paths carry the owner:
    `specs/design/components/<name>/…`, `specs/design/dependencies/<name>/…`.
11. **Changes only, no full manifest** in this cut.
12. **One request.** Both answers are wanted by the same click, so `get-build-preflight` carries
    both. Rejected: a second op, which makes one click wait on two requests.
13. **The action says what happens** — `Build v3` / `Rebuild v2`, never *Cut v1 & build*.

14. **An untouched name field claims nothing.** The request carries a version only when the user
    replaced the suggestion. So the suggestion stays a suggestion: a build that loses the race for
    `v3` is given `v4` rather than a conflict, while a name somebody typed is never turned into a
    different one. The two rules meet in one place — the platform re-suggests only what it
    suggested.
15. **A version is recognised by its annotation, not its name.** Every spec tag carries the subject
    `Spec <name>`, which is what keeps a release tag (or a legacy `v<N>-<M>` design tag) from being
    read as a version once names are free.

Candidates for ADR graduation (flow step 4): decision 5–8 as one ADR (it amends the PRD's **Spec
versioning** section), and decision 1 as a console convention.

## Contract changes

`packages/contracts/api/v1/openapi.yaml`:

```yaml
BuildPreflight:
  properties:
    needsInput: { type: boolean }
    needsResolution: { type: boolean }
    items: { type: array, items: { $ref: '#/components/schemas/PreflightItem' } }
    # new
    currentVersion:
      type: string
      description: The newest version's tag name, "" when the project has none.
    suggestedVersion:
      type: string
      description: What the version box is prefilled with — v<count+1>, incremented until free.
    specUnchanged:
      type: boolean
      description: The specs/ tree at HEAD matches the newest version's, so a build rebuilds it.
    changes:
      type: array
      description: What this version changes, against the newest version. Empty when specUnchanged.
      items:
        $ref: '#/components/schemas/BuildChange'

BuildChange:
  properties:
    name: { type: string }
    kind: { type: string, enum: [component, external, platform-resource, requirements] }
    state: { type: string, enum: [new, changed, removed] }

BuildRequest:
  properties:
    inputs: { ... }
    # new
    version:
      type: string
      description: The tag name to cut. Empty takes the suggested one. A name already in use is a 409.
```

## States to design

- **Loading** — Build pressed, the commit and preflight in flight (the button already carries this).
- **Empty change list** — an unchanged spec tree: *no spec changes since `v2`*, the version box
  locked, the action **Rebuild v2**.
- **First build** — no previous version: the heading reads *What this version creates* and every row
  is `new`.
- **Invalid name** — a character a git tag cannot hold: inline rule under the box, Build disabled.
- **Taken name** — inline error, Build disabled; the server's 409 for the race.
- **Preflight failed** — the existing build error surface; no dialog opens.
- **Gate refusal (422)** — the existing checklist, rendered over the build dialog.

## Open questions

None.

### LINKED ISSUE #751: aep-api: a version carries the name the user gives it, and preflight says what the build changes
The frontend half of [#749](https://github.com/wso2/labs-agentic-engineer/issues/749) is built, mock-validated and open as [#750](https://github.com/wso2/labs-agentic-engineer/pull/750). This is the `aep-api` work it waits on.

**Feature branch: `feat/749-build-dialogs`.** The backend lands there, not on `main` — the two halves meet on the branch and only the branch merges.

## What the console now asks for

One Build click opens one dialog. Which one, and what it says, comes entirely from `get-build-preflight`, which the click already waits on:

- the open dependencies it already returns (unchanged),
- the newest version's **name**, the **suggested** name for this one, whether the spec tree is **unchanged**, and the list of what this version **changes**.

And the build request now carries the name the user settled on.

## Contract (already on the branch)

```yaml
BuildChange:               # new
  name:  string            # component / dependency / resource name, as the design writes it
  kind:  component | external | platform-resource | requirements
  state: new | changed | removed

BuildPreflight:            # + four fields
  currentVersion:   string   # newest version's tag name; "" when never built
  suggestedVersion: string   # v<count of versions + 1>, incremented until free
  specUnchanged:    boolean  # specs/ at HEAD matches the newest version's tree
  changes:          BuildChange[]   # empty when specUnchanged; all `new` on a first build

BuildRequest:              # + one field
  version: string          # the tag to cut; empty takes the suggestion; in use → 409
```

Full diff: `git diff main -- packages/contracts/api/v1/openapi.yaml` on the branch.

## The behaviour behind it

**A version carries the name the user gives it** ([ADR-0030](https://github.com/wso2/labs-agentic-engineer/blob/feat/749-build-dialogs/apps/console/design/decisions/ADR-0030-a-version-carries-the-name-the-user-gives-it.md)). Three consequences for `spec`:

1. **`v<N>` stops being identity.** `latestRequirementsTagInfo` takes the maximum of `^v(\d+)$`, and `LatestSpecTag` and the unchanged-detection both ride on it. With free-text names those reads must order by the tag's **creation time** — `%(creatordate:iso-strict)` is one more field in the `for-each-ref` that `gitfs.readTags` already runs, and one more field on `TagInfo`. Creation time also separates two tags on one commit, which ancestry cannot.
2. **A supplied name is used verbatim, and a collision is an error.** `createAnnotatedTag` heals a conflict today by recomputing the next number — that only ever worked because the number advanced on its own. A user-supplied name that is taken must come back as a conflict the handler maps to **409**, never a silently different tag.
3. **The suggestion is `v<count of versions + 1>`**, incremented until free. It is only a default; the field may be replaced.

**A name labels a snapshot, it does not make one.** Unchanged detection is unchanged in spirit: when HEAD's `specs/` tree matches the newest version's, the build still reuses that version and reopens its milestone. `version` is ignored on that path, and `specUnchanged: true` is what tells the console to lock the field and say *Rebuild*.

**The change list** is the diff between the newest version's tag and HEAD, mapped onto owners by path — `specs/design/components/<name>/…` → `component`, `specs/design/dependencies/<name>/…` → `external` (or `platform-resource` where the design says so), the PRD → one `requirements` row. A path under a directory that is new to this version is `new`; one whose directory is gone is `removed`; anything else `changed`. On a first build every row is `new`.

## Notes

- Preflight is the only read the click makes, and that stays true — no second op.
- Nothing about a version name reaches OpenChoreo: the tag names the spec snapshot and the milestone, not a release.
- The older per-artifact `v<N>-<M>` design tags stay legacy and untouched.

https://claude.ai/code/session_011VNg9Gdpd7SMv6c4VD6PLn

hevayo:
Both halves of #749 are on this branch and were validated together on the local setup, so this is the whole feature.

## What changed for the user

One Build click opened one of two different containers — a right-hand drawer when a dependency was unresolved, a centred modal otherwise. Neither said what the build would do, and the version was always `v<N>`.

Now the click opens exactly **one dialog**, and both are dialogs ([ADR-0029](../blob/feat/749-build-dialogs/apps/console/design/decisions/ADR-0029-build-asks-in-a-dialog-that-lists.md)):

**Resolve dependencies** — the names of the dependencies with no provider or no interface on file, and one action, **Resolve**, which runs the guided flow over all of them. No per-row button: the rail already reaches every definition.

**Start build** — the version's **name** first, then **What changed since `<version>`**, grouped so a name says what it is:

```
  Components
    orders-api
    reports-web                new
  External dependencies
    currency-service
    legacy-mailer              removed
  Platform resources
    postgres-cnpg              new
```

The two dependency kinds are never one group: the platform stands one up for you, the other needs a provider and its keys. A removed row states a fact rather than promising an act — a build deprovisions nothing, so its resource stays. The requirements are not a row at all: every row names something that exists once the version is built.

**The name is the user's, and it is the tag that gets cut** ([ADR-0030](../blob/feat/749-build-dialogs/apps/console/design/decisions/ADR-0030-a-version-carries-the-name-the-user-gives-it.md)). The field suggests `v<count of versions + 1>`, refuses what a tag cannot hold, and names a collision before the request goes out. An untouched field claims nothing — the suggestion stays a suggestion, so a build that loses the race for `v3` is offered `v4` rather than refused, while a name somebody typed is never turned into a different one. An unchanged spec tree still reuses its version: field locked, action **Rebuild `<name>`**.

## What it took server-side

`v<N>` was doing three jobs — identity, order, and the name a person reads. The name is now the user's, and the other two split:

- **Identity** is the tag name itself, whatever they typed.
- **Order** is the tag's creation time — one more field in the `for-each-ref` the mirror already runs, which also separates two versions naming the same commit.
- **Which tags are versions** is answered by the annotation: every spec tag carries the subject `Spec <name>`, so a release tag, or a legacy `v<N>-<M>` design tag, is not mistaken for one.

A supplied name is used verbatim; a collision is `ErrVersionNameTaken` → **409**, never a quietly different tag. Only a name the platform suggested is recomputed past a racing pusher.

Preflight answers the same click's other half — current name, suggestion, whether the tree moved, and the change list, computed from two `specs/` tree listings (platform-resource rows come from the designs, since a resource owns no directory). A version read that fails degrades the dialog rather than refusing the build.

## Contract

```yaml
BuildChange:               # new — name · kind (component|external|platform-resource) · state (new|changed|removed)
BuildPreflight:            # + changes[], currentVersion, suggestedVersion, specUnchanged
BuildRequest:              # + version — the tag to cut; empty takes the suggestion; in use → 409
```

The new `kind`/`state` enums made the generator disambiguate three existing constants (`gen.Resource` → `gen.WorkloadDependencyDTOKindResource`, same for `OrgService`); one test referenced them and follows.

## Proof

**Mock mode** — `aep:mock:build` switches `changes` / `blocked` / `rebuild` / `first`; screenshots on the thread.

**Real git** (`TestBuildVersionFactsOverRealGit` and three siblings) — the spec harness's bare origin plus the production workspace engine: real annotated tags with real creation dates, a real directory deletion. Asserts `changed` / `new` / `removed` / a platform resource arriving, that creation time beats a name that sorts above it, that a release tag is not a version, and that the suggestion steps past a name taken by hand.

**The local setup**, both images built from this branch:

```
small-shop-handmade2   (no versions)  suggestedVersion v1 · 8 rows, all new · needsResolution true
small-shop-handmadev6  (holds v1)     currentVersion v1 · suggestedVersion v2 · specUnchanged true
  + probe: a file under requirements and a new dependency directory
    → Requirements not listed · tax-service new     (then reverted; the spec ends byte-identical)
```

That run is what caught the one real bug in here: the preflight **handler's response mapping** never carried the four new fields, so the service computed them and the JSON dropped them, with nothing failing anywhere (`0996fcb0`). Mock mode cannot see that class — MSW answers from its own fixtures. Two tests now assert the decoded HTTP response rather than the service.

Gates: full Go suite · 1857 console tests · `golangci-lint` 0 issues · `deadcode-check` · `gen-api-check` · `license-check` — all clean.

## Also in here

- `BuildDependencyDrawer` and `buildScope.ts` deleted; ADR-0028's Build-drawer paragraph marked superseded.
- PRD **Spec versioning** no longer describes an incrementing `v<N>` sequence as the model, and the feature inventory carries the entry.
- `design/lexicon.md` — the Build drawer entry replaced by the two dialogs' copy.
- `latestRequirementsTagInfo` went with the number-ordering it existed for.

## Known, not fixed here

The change list shows dependency definitions no component references — five orphaned print services on `small-shop-handmade`. The design is what is wrong, not the dialog: #752 tracks the agent minting a new definition on reconsider instead of editing the one it has.

Closes #749
Closes #751

https://claude.ai/code/session_011VNg9Gdpd7SMv6c4VD6PLn


hevayo:
**Tried on the local setup** — both halves rebuilt from this branch into the running stack (`deployments-aep-api`, `deployments-console`), then the real `get-build-preflight` walked against real projects.

It caught one bug mock mode structurally could not: the preflight **handler's response mapping** never carried the four new fields, so the service computed them and the JSON dropped them — an empty name field and no change list, with nothing failing. Fixed in 0996fcb0, with two tests that pin the response itself rather than the service.

Live results after the fix:

```
small-shop-handmade2   (no versions yet)
  suggestedVersion v1
  changes  Requirements new · ceramics-admin new · ceramics-api new ·
           ceramics-storefront new · currency-provider new · email-provider new ·
           payment-provider new · ceramics-db new · user-auth new
  needsResolution true   → the Resolve dependencies dialog (currency-provider)

small-shop-handmadev6  (holds v1)
  currentVersion v1 · suggestedVersion v2 · specUnchanged true · no changes
  needsResolution false  → Start build, field locked, action Rebuild v1
```

Worth noting what those rows prove: `ceramics-db` and `user-auth` are **platform resources**, which own no directory in `specs/` — they came from reading the designs at both trees, which is the only way a resource can be seen arriving.

Not yet seen live: `changed`/`removed` rows on a second version, which needs a real spec edit and a real cut (an agent run). Unit tests cover them, and mock mode shows the rendering.

The stack is currently running this branch.

hevayo:
**changed / removed rows, tested two ways.**

**Over real git** (8475d2c3) — the spec harness's bare origin plus the production workspace engine, so: real annotated tags with real creation dates, real trees, a real `git` deletion. `TestBuildVersionFactsOverRealGit` seeds a spec, tags `v1`, then edits the PRD, adds a component, deletes a dependency directory and gives a component a platform resource, and asserts exactly:

```
Requirements  changed
orders-api    changed
reports-web   new
legacy-mailer removed     (external — the deleted directory)
orders-cache  new         (platform-resource — declared in a design, no directory)
```

…and nothing for `README.md`, which is not part of the spec. Three more hold the ranking: creation time beats a name that sorts above it (`payments-v2` after `v9`), a release tag is not a version, and the suggestion steps past a name taken by hand.

**On the live stack**, against `small-shop-handmadev6` (a real project holding `v1`):

```
before    specUnchanged true · currentVersion v1 · suggestedVersion v2   → Rebuild v1
edit      + specs/requirements/_probe.md · + specs/design/dependencies/tax-service/
after     changes: Requirements changed · tax-service new                → Build v2
revert    both probe files deleted
restored  specUnchanged true                                             → Rebuild v1
```

Only files the probe created were touched, and the project's spec ends byte-identical to where it started. A live `removed` row would mean deleting a dependency the project actually authored and restoring it byte-for-byte afterwards, so that one is left to the git-level test rather than run against someone's project.

hevayo:
Worked through the CodeRabbit findings — four applied in `634a9e0e`, one declined.

**Applied**

- **`version_naming.go` — arbitrary `v<N>` tags classified as versions** (the major one, and the right catch). The rule took any `v<N>` NAME, so a repo's own release tag called `v1` could become the newest version — the baseline the change list is computed against and the tree the rebuild check compares. The annotation decides now: `Spec <name>` as this platform writes it, or `Requirements <N>` from the retired per-artifact save, whose tags really are this platform's versions. A name speaks only for a tag carrying no annotation at all, which is the lightweight-tag case the fallback existed for. Tests: an annotated `v2` named like a version and subjected *"ship it"* is refused (unit and over real git), the legacy subject is accepted, and the unannotated fallback still works.

  Worth noting what this shook out: **seven fixtures were tagging with prose** (`"spec v1"`, `"oob tag"`, `"external v1"`) where production writes `Spec v1`. They passed only because the old rule fell back to the name. They now write what production writes.

- **`SpecView.tsx` — turn-end refresh could reopen a closed dialog.** Real: unsubscribing does not stop a promise already in flight, so a turn ending as the user closes the dialog would set it open again over the conversation they just went back to. The cleanup marks the flight stale.

- **`openapi.yaml` — undeclared 409.** The `version` field's description promised a conflict the operation never declared. Declared, with what it means (name in use, or a build already running).

- **`build_version_test.go` — assert the change row's name.** Fair: `Name`, `Kind` and `State` map independently, so a dropped `Name` would have passed. Asserted.

**Declined — "preserve explicit version-name submission after an edit"**

The suggestion is to track whether the field was edited, so that typing a value equal to the suggestion submits it explicitly rather than as "take the suggestion".

The two cases are not distinguishable in effect. `suggestedVersion` is computed server-side as a name that is FREE — it counts the versions and steps past any taken name — so a collision on it is only reachable through a race with another build. In that race the two behaviours differ by exactly one thing: re-suggestion hands the user `v4`, an explicit submission hands them a 409 over a name they never chose deliberately (they accepted an offer). We picked re-suggestion for that case on purpose, and it is decision 14 in #749.

Adding an `edited` flag would add state whose only observable effect is to make the accepted-default path fail in a race. Skipping it.

xlight05 (approved):

PR #761, opened 2026-09-12
fix(aep-api): the Thunder token cache trusts the wall clock and recovers from a 401

xlight05:
## What broke

Clicking **Build** on a project answered **"publisher credentials unavailable"**. aep-api's log:

```
publisher provision for build failed: idp_service.EnsureOrgPublisher: findApp "aep-publisher-default":
thunder list apps returned 401: {"code":"AUTH-4010", ... "Authentication is required to access this resource"}
```

A token minted by hand from the same client (`aep-system-client`, `scope=system`, resource indicator set) listed applications with 200, so the credentials were fine. aep-api was presenting a token Thunder had already expired, and kept presenting it.

## Root cause

`thundersvc` cached the system token with `tokenExpiry = time.Now().Add(expires_in − 30s)` and checked `time.Now().Before(tokenExpiry)`. Go evaluates that comparison on the **monotonic** clock. The laptop slept on battery between 16:29 and 16:54 IST; the Colima VM was paused with it, its monotonic clock stopped, and the wall clock (which Thunder judges `exp` by) stepped forward on resume.

Measured on the running aep-api container after the wake:

| clock | age of the process |
|---|---|
| monotonic (`/proc/uptime − starttime`) | 2990 s |
| wall (`docker inspect StartedAt` → now) | 4198 s |

About twenty minutes lost. Thunder's access log shows the token minted at `10:48:33Z` (the 1030-byte response is the system client with the resource indicator; the org-unit lookup follows it), expiring at `11:48:33Z`. The failing `GET /applications` came at `11:49:47Z` with **no token mint before it**: the cache believed nineteen minutes remained. And because a 401 never touched the cache, every admin call would have failed until the monotonic deadline came round.

Two fingerprints that separate this from the earlier resource-indicator trap: Thunder answers **401** for an expired or invalid token and **403** for a token without the `system` scope. The existing mint-time scope assertion is untouched and still passes.

## The fix

In `services/aep-api/internal/clients/thundersvc`:

- **The cache trusts the wall clock.** Expiry is kept as a Unix second taken from the JWT's own `exp` claim (falling back to `expires_in`, then a minute) and compared through an injectable `now`. The 30 s refresh skew stays; a token already inside the skew is not reused.
- **A 401 evicts and retries once.** Every admin request (client.go and directory.go) goes through one send helper. On 401 it drops that token from the cache if it is still the cached one, mints once, replays the request with the fresh token (bodies via `GetBody`), and returns that response. A second 401 is returned as the answer and the fresh token is not kept. The token endpoint itself stays outside the helper.

## Tests

New `token_cache_test.go`, against the existing Thunder mock (which gained mint counting, an `exp` claim and token revocation):

- the wall clock jumps two hours past expiry → a fresh token is minted
- `exp` ten seconds out while `expires_in` says 3600 → the token is not reused
- Thunder revokes the cached token → the call recovers with exactly one extra mint and the cache holds the fresh token
- Thunder refuses every token → exactly one retry, the 401 surfaces, nothing stays cached

All four fail against the previous code. `go test` for the package and for every consumer package (`organization`, `identity`, `app`, `clients/...`) is green; build, vet, golangci-lint (0 issues) and the Go dead-code gate are clean.

## Proof from the live environment

The diagnosis above is from the local converged cluster on 2026-09-11 (Thunder access log, aep-api log, the container clock measurement, and the hand-minted token that succeeded). The fixed image has **not** been deployed yet; the running stack recovered on its own once its monotonic deadline passed.

### DIFF
--- a/services/aep-api/internal/clients/thundersvc/client.go
+++ b/services/aep-api/internal/clients/thundersvc/client.go
@@ -188,10 +188,17 @@ type client struct {
 	systemAud  string
 	httpClient *http.Client
 
-	mu          sync.RWMutex
-	cachedToken string
-	tokenExpiry time.Time
-	tokenSfg    singleflight.Group
+	// now is the clock the token cache reads. It is the WALL clock on purpose:
+	// the cache compares Unix seconds, never time.Time values, because Go's
+	// time.Before prefers the monotonic reading and a VM paused with a sleeping
+	// laptop stops that clock while the wall clock, and Thunder's `exp`, move
+	// on. Tests substitute it.
+	now func() time.Time
+
+	mu              sync.RWMutex
+	cachedToken     string
+	tokenExpiryUnix int64 // wall-clock second after which the cache stops trusting cachedToken
+	tokenSfg        singleflight.Group
 
 	// Default OU id, looked up once on first EnsurePublisherApp call
 	// and cached. Thunder's UI nests every org under a root OU
@@ -220,6 +227,7 @@ func New(cfg Config) Client {
 		systemSec:  cfg.ClientSecret,
 		systemAud:  cfg.SystemResourceIdentifier,
 		httpClient: hc,
+		now:        time.Now,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


xlight05:
Superseded by #825, which landed the same fix in 99718e12 (`fix(aep-api): judge the Thunder system token by the wall clock, and evict it on 401`): the cache expiry is judged by the wall clock against the token's own `exp`, and a 401 to the cached bearer evicts it.

The one piece not carried over is the single retry: on main the call that hits the 401 still fails, and the next call mints a fresh token. If that turns out to matter, it can be a small follow-up on main's `tokenEvictingTransport`.

PR #763, opened 2026-09-12
feat(scopes): phase 2 — resource servers, project roles, bindings, ticket

xlight05:
## Scopes phase 2 — directory: resource servers, roles, bindings, ticket

Design of record: https://claude.ai/code/artifact/372addfc-556c-4d8d-a693-f74ea2266bbd (Version 3). Plan: `docs/design/draft/generated-app-scopes-impl-plan.md` §3 Phase 2 (draft, not committed). Builds on #762 (phase 1).

A build on the environment directory now leaves the resource-server tree, the project's roles, their group assignments and the client allowlist exactly as the tagged `security.json` declares, and the roles-gate ticket carries scopes.

### What lands

- **Directory port** (`internal/identity/ports.go`): 15 concept-level verbs for resource servers, resources, actions, roles and assignments (`DirectoryID` opaque, `Principal{group|user|app}`); `ResourceServerIdentifier(org, project)` = `https://aep.wso2.com/orgs/<org>/projects/<project>` (the token `aud`, the gateway audience, the SPA `resource`); `RoleName` = `<project>/<Role>`.
- **thundersvc** client for `/resource-servers`, resources, actions, `/roles`, assignments, with the Thunder 1.0.0 status codes measured in the spikes mapped to sentinels (RES-1004/1013/1014/1006, ROL-1004/1012, SAZ-4030).
- **Store**: `idp_resource_servers` (PK org, env, project; identifier unique per directory) and `idp_role_bindings` (one row per role×group; an empty group means "assigned to nobody"). `ReplaceRoleBindings` converges in one transaction. No data migration; both tables are new.
- **Ensure passes** (`internal/identity/ensure.go`), in the design's order: classify (an `assignTo` that neither `groups[]` nor the directory holds fails by name before any write) → accounts → groups (additive, enrolment) → resource server (find by identifier, delimiter `:`) → resources/actions (converge to the tag: creates before deletes, leaf-first; a description change is an update, never a delete, so no grant is ever cascaded away by a reworded sentence) → roles (one PUT per role only when the permission set differs; group principals converge to `assignTo`; user and app principals are never removed; roles the tag dropped are unassigned then deleted). **A rebuild of the same tag writes nothing** (asserted on the fake's call log).
[...]

PR #764, opened 2026-09-12
feat(scopes): phase 3 — gateway and service enforcement

xlight05:
## Scopes phase 3 — gateway and service enforcement

Design of record: https://claude.ai/code/artifact/372addfc-556c-4d8d-a693-f74ea2266bbd (Version 3). Plan: `docs/design/draft/generated-app-scopes-impl-plan.md` §3 Phase 3 (draft, not committed). Builds on #763 (phase 2).

A deployed API now answers as the design's matrix says: the gateway enforces one scope per operation rendered from the component's OpenAPI contract, and the service enforces the same scope from the same file when the gateway is bypassed.

### What lands

- **Trait** (`deployments/manifests/api-platform/api-configuration-trait.yaml` + the byte-identical helm copy, with `deployments/scripts/check-trait-copies.sh` wired as `make manifests-check`, a prerequisite of `make test`): `operations[]` items carry `method`, `path`, optional `public`, optional `policies[{name: jwt-auth, version: v1, params: {scopes: {anyOf: [one handle]}, audiences}}]`; `claimMappings` gains `scope → x-user-scopes`; API-level jwt-auth is suppressed once any operation carries its own policy or `public`; no `onFailureStatusCode` (rejected at admission and inert cluster-wide); `apiVersion` `gateway.api-platform.wso2.com/v1`. The CEL tests `policies` before `public`, so a row carrying both keeps its scope. The header documents the measured facts and the decision not to gate on groups (scopes are the one authority).
- **Operations projection** (`internal/spec/openapi_operations.go`, `internal/projects/api_operations.go`): the gate's own security reader is reused, so the write gate and the gateway cannot disagree about a `security` block. Absent → signed-in (a jwt-auth policy with no `scopes`, never `anyOf: [openid]`); `[]` → `public: true`; one scope → `scopes.anyOf: [scope]`. One `OPTIONS` row is synthesised per path (an undeclared method 404s before any policy, so every cross-origin preflight would die). Anything the gate refuses the projection refuses; HEAD/TRACE rows are skipped with a note rather than refused (refusing dropped the whole table to `/*`). Components without the sign-in dependency keep `/*` with API-level jwt-auth; the audience is pinned per environment to the project's resource-server identifier only for sign-in components.
- **Gate hardening**: the document-level `security` default must be exactly `[{oauth2: []}]` on both gates; a default naming a scope previously passed and, with the projection, would have rendered every inheriting operation as merely signed-in.
- **Stack assets** (verbatim copies from the spike): `skills/go/assets/scopes_middleware.go` (exactly-one-scope from the generated `Oauth2Scopes` context; `X-User-Id` empty → 401; missing scope → 403 + `WWW-Authenticate: Bearer error="insufficient_scope", scope="<handle>"`; `hasScope`; ownership idiom), `skills/ballerina/assets/scopes.bal` + `scope_table_drift_test.bal` (transcribed table + drift test; verify step is `bal build && bal test`).
- **Skills**: `go` (contract-first step with pinned oapi-codegen v2.7.0; `RequireScope` in `ChiServerOptions.Middlewares`, never `r.Use()` — fail-open, silent; three visibility values; both Dockerfile branches), `ballerina`, `api-management` (header table with `X-User-Scopes` as the authority, "the spec is the gateway config", trust boundary, 404-not-403), `thunder-authentication` backend half (SPA half is phase 4), `aep-validation` (judge grants before signing in; the gateway answers 401 for every failure, the service 403; sign in fresh after a grant change), `aep` wiring reference (a web app's own endpoint visibility; the `resource` output and `validityPeriod`).
- **`deployments/scripts/verify-scopes.sh`**: the regression probe. Mints the system token and each test user's token headlessly through Thunder's flow API (five calls; no browser), asserts `Programmed=True` and the `Programmed` condition's `observedGeneration` before the matrix, and runs it at the gateway (200 with scope, 401 no token, 401 platform-IdP token, 401 missing scope with a body byte-identical to the no-token case, 204 preflight, 200 public, 200/401 signed-in class, 404 undeclared) and in-cell against the service (401 without `X-User-Id`, 403 + `insufficient_scope`, 200 with; ownership row counts). No credential touches argv or disk.

### Verification
V-unit / V-int (all green): aep-api build/vet and spec/projects/securityspec/dependencies/identity/app tests; `deadcode-check`, `gen-api-check`; `@aep/agent-stream` 258 tests, `gen` idempotent, all three vendored artifacts byte-identical; 75 skill node tests; the skill-library load test; all 29 skill frontmatters parse; `check-trait-copies.sh` and `make manifests-check`; server dry-run of the ClusterTrait and the example RestApi (`configured` / `created`); the example's 15 operation rows equal the projection golden row for row; the Go asset vetted and the Ballerina assets built and tested in throwaway packages; root `deadcode-ts-check`/`typecheck`/`lint`/`license-check`; gofmt clean.

V-live (local plane; transcript with the rendered RestApi, the verify-scopes table, header dumps in the session scratchpad) — **PASS on every item; 19 PASS / 1 SKIP in `verify-scopes.sh`**:

- aep-api rebuilt and proven running; the new ClusterTrait applied; skills synced to the org library. The coding run started and copied `scopes_middleware.go` verbatim from the phase 3 skills (proof the skills reach the runner), then died on the upstream API quota; the fixture was deployed instead from the hand-written Expense Tracker app through the platform's own PR → auto-merge → `EnsureComponent` → image build → deploy path, so the RestApi below is rendered by this branch's projection.
- Rendered RestApi `scopes-p3-expense-expense-api-…-http`: 15 rows; `GET/OPTIONS /claims` → `claims:read`, `POST /claims` → `claims:submit`, approve/reject with their handles, `/reports/monthly` → `reports:read`, `GET/OPTIONS /me` → jwt-auth with no `scopes` key, `GET/OPTIONS /health` → no policy; every policy carries `audiences: [https://aep.wso2.com/orgs/default/projects/scopes-p3-expense]` and `claimMappings.scope: x-user-scopes`; `spec.policies` is `[cors]` only; `Programmed=True`, condition `observedGeneration == generation`.
- Gateway: 200 with scope; 401 no token; 401 platform-IdP token; 401 environment system token; 401 missing scope with a body byte-identical to the no-token case and no `WWW-Authenticate`; **204 on the synthesised `OPTIONS` row although it carries its sibling's jwt-auth with scopes** (preflight short-circuits before jwt-auth, as the trait's design note claims); 200 `/health`; `/me` 200 with either token and 401 without; 404 undeclared.
- Service, in-cell: 401 without `X-User-Id`; 403 + `WWW-Authenticate: Bearer error="insufficient_scope", scope="claims:read"`; 200 with the scope; ownership: employee sees 2 rows, approver 8.
- Headers at the service: `x-user-scopes: openid profile email group ou claims:read claims:submit` (raw, OIDC scopes included); `x-user-groups: ["Employees"]` (JSON) and `""` for a user in no group; `Authorization` stripped and re-presented as `x-forwarded-authorization`.
- SKIP: the two-group `x-user-groups` shape (no two-group account exists on this plane).

Fixes the live run forced, each committed with a test: `ce478334` the SPA's nginx `/api` proxy now clears `X-User-Scopes` (plan Δ P6 §7.27 had slipped; a forged header through the proxy now answers 401); `ba25fbc0` `verify-scopes.sh` requests the scopes it asserts (Thunder returns the intersection of requested and granted, not every grant); `fc6c481e` the script asks the directory for the catalog rather than the gateway's operation table (`claims:read-all` guards no operation).

Findings, not fixed here: `issuers: []` on every policy because `resolveIssuers` returns nil for a `kind=platform` org IdP profile (the audience pin alone rejects foreign tokens; pre-existing); a new `claimMappings` entry takes ~60 s to reach the policy engine after `Programmed=True` (design note 7 understates it); two delivery-layer defects outside this phase: a run whose agent pod dies can never land (re-dispatch onto a closed cycle), and a rebuild of an unchanged spec never reaches the deploy stage.

### Interim state and notes
- A `HEAD`/`TRACE` row a contract declares is absent from the gateway's operation table rather than refused (neither method is in the RestApi CRD's set); reported on `DesiredDeployment.APIOperationsNotes` and logged.
- `security.json` is validated at three points with two severities by design: `POST /files/apply` reports defects as non-blocking warnings, while `SaveSpec` and the design-save gate refuse, so no tag can be cut over a broken document.
- Nothing yet gates an operation becoming public: `security: []` renders `public: true` with no cross-check against the screens or roles that reach it. A phase 5 warning rule is proposed.
- The `.bal` assets carry no Apache header (`.bal` is outside the license check's match set); the `.go` asset does because the check demands it. Whether generated apps should carry the platform's header is an open question.
- On this cluster the RestApi's `observedGeneration` is on the `Programmed` condition, not on `status`; the spike notes and the plan say otherwise.


xlight05:
Superseded: reopened as a GitHub stacked PR with the same branch pushed to wso2/labs-agentic-engineer (fork-headed PRs cannot join a stack). Body and history carry over to the new PR.

PR #778, opened 2026-09-15
fix(authz): Implement RBAC for AE users 

senithkay:
## Summary

`aep-api` had no per-endpoint authorization. `tenantGate` bound the caller's organization and nothing after it asked what that member was allowed to do, so any authenticated org member could call any operation the contract exposed. The console shipped permission-gated UI against a hardcoded stand-in set — gating a caller could bypass by calling the API directly, because nothing on the server agreed with it.

This PR builds the AE (`ae:*`) permission model end to end: the vocabulary, the roles, the IdP configuration that makes a permission reachable, a deny-by-default gate in front of every contract operation, and the console changes that make a denial explicable rather than a dead end.

## How a permission travels

```
Thunder role (ae-admin / ae-developer)
  → OAuth scope claim on the access token        ADR-0027
    → auth.Claims.Permissions()  (aep-api)       permission_gate.go
    → permissionsFromScope()     (console)       auth/permissions.ts
      → AE→OC bridge provisions an OC AuthzRole  internal/authz
```

The same key is the unit at every hop. No role names are checked anywhere.

### Vocabulary and roles

14 permission keys in `authz.Permission`, the single typed source shared by the role catalog, the AE→OC action catalog, and the gate. Two roles: `ae-admin` (all of them) and `ae-developer` (requirement-view plus the design and build pairs). The console mirrors the list by hand until the backend exposes it as a generated contract type.

### The AE→OC bridge

`internal/authz` provisions OC `AuthzRole` / `AuthzRoleBinding` CRs so OpenChoreo can authorize calls the platform makes on a user's behalf, entitled on the `groups` claim. `EnsureAuthzRole` runs as onboarding's first blocking step — and under the BFF's service identity, since a first-time user has no OC grant yet and cannot bootstrap their own role under their own token. A Helm-chart claim mismatch (`sub` vs `client_id`) that would have silently no-op'd that on non-local installs is fixed alongside.

### Making a permission actually reachable

Two findings from live cluster testing, both invisible in code review:

- Thunder narrows a requested scope to what the caller's role holds, but never grants what was not requested. The console must request **every** `ae:*` scope or the omitted one can never be granted to anyone, however privileged.
- An `ae:*` scope requested without an RFC 8707 `resource` indicator resolves against the platform's default resource server, silently dropping the scope *and* rewriting the token's audience — which then fails `aep-api`'s own audience check. The console now sends `resource`, and `JWT_AUDIENCE` accepts that identifier.

Thunder resources (`ae` resource server, actions, groups, roles) are declared for every topology: `aectl`, the single-cluster bundle, and `dev-thunder-setup`.

### The gate

`internal/edge/permission_gate.go`, deny-by-default, beside `tenantGate` and running after it. Every contract operationID appears in exactly one of `operationPermissions` or `permissionGateCarveOuts`; `TestPermissionGateCoverage` fails the build otherwise, and a mirror test rejects a carve-out naming an operation the contract no longer has. **65 gated, 23 carved out.**

Two rules are worth stating because they are what most of the gate's churn was about:

- **Entry to a surface is gated on its view permission exactly.** A write permission authorizes mutations; it never admits its holder to a page. `ae:build` does not satisfy a row reading `ae:build-view`, even though both built-in roles that hold one hold the other — the gate describes what a permission *means*, not which roles happen to exist today.
- **Where an OR appears it spans features**, never a write permission and its own view sibling — `GET /config`'s two credential cards, read by either credential permission and then redacted per section, are the example.
- **An OR is also wrong when two features differ by path rather than by operation**, because it grants each one's permission everywhere the other reaches.

That last rule came out of review. The AI chat panel is one component mounted twice — the project spec chat and the marketplace registration assistant — and was OR-gated on `ae:design` OR `ae:resource-config`. But a turn's `instruction` carries `/<skill>` flow commands verbatim for the server to expand, so a caller holding only `ae:resource-config` could send `/design` at a **real** project and get exactly what `generate-design`'s `ae:design` gate exists to control. The dedicated operation had made the console's button safe while leaving the generic one open. Those seven rows now require `ae:design`, and the marketplace pseudo-project is resolved per request instead.

So three operations can't be decided by a flat row: `UpdateConfig` (keyed on which body sections the patch touches — an AND, not an OR), the chat panel (keyed on project), and the spec collaboration room, which can't be settled at the gate at all; see below.

### The collab room

Joining a spec room needs `ae:design-view`; changing it needs `ae:design`. Only the first was enforced. The console hid its editing controls from a viewer, but that is a property of our client, not of the room — a read-only participant's socket accepted Yjs updates like anyone else's, so the document could be edited by someone with no design-write permission through any client that does not run our code. `ApplyFiles`' gate cannot catch it: it sees whose *token* makes the request, never whose edits are in the document.

`validate-collab-access` now answers `canWrite`, and `services/collab` acts on it twice — marking the connection read-only so Hocuspocus drops that socket's updates, and keeping the participant's token out of the commit path.

The second half is not only a permission matter. The room's `lastToken` authenticates the unload flush and was simply the most recent joiner's; once `ApplyFiles` required `ae:design`, one viewer arriving would have become the credential every later flush used, failing all of them and stranding the real authors' uncommitted work. A viewer can still trigger a flush — it commits as the room's last writer instead.

### Console denial UX

A 403 the user cannot act on is a bug even when the gate is right. The Spec page's empty state, the Overview's Architecture panel and `useConversationLog` each showed a generic failure for what was really a permission problem, or fired a doomed request at all; each now checks first and explains. Validation gained the `PermissionRestrictedPage` guard Builds and Deployments already had. The Overview track's legs and the sidebar's nav items now lock on the exact permission their destination gates on, closing a "leg says open, page says no" dead end.

The spec editor is genuinely read-only without `ae:design`: Tiptap's `editable` flag only blocks typing, while the toolbar, accept/reject bar and mermaid controls dispatch transactions programmatically — all are hidden, not disabled. That is the UX half of the collab fix above, not a substitute for it.

### Gaps closed rather than papered over

- `POST /authz/role-permissions` rewrote the org's OC AuthzRole actions and required no permission to do it. Nothing called it. The route is removed; `AuthZService.ModifyRolePermissions` stays as unwired infrastructure for the operator surface that has not been designed.
- `PATCH /config {idp}` was ungated, and is not inert: the deploy path reads the org's IDP profile and pins the gateway's `jwtAuth` issuers to it, and a kind switch deletes the org's Thunder publisher app. No AE permission describes identity configuration and nothing writes the section, so the gate refuses it outright rather than inventing a permission for a surface that does not exist.
- `GET /config` leaked `codingAgent.updatedBy` — the field documented as the compensating control for that endpoint's coarse RBAC — to a caller who cannot write the section. Now redacted with it.
- Two callerless IDP routes (`rotate-idp-client-secret`, `discover-idp`) are retired the same way: route out, service kept.
- `DirectOCStrategy` resolves a tokenless, unmarked OC call to the BFF's full-privilege identity. The webhook and dispatch paths depend on it and `tenantGate` keeps console traffic away, but it is a fail-open; it now logs as one, pending the audit that lets it become a refusal.

### Build reproducibility

CI's contract-freshness gate failed on this branch reporting a stale contract, and nothing was stale — the same commit passes `make gen-api-check` locally. `server_gen.go` embeds the contract as a gzipped, base64-encoded blob for `GetSpec()`, gzip output is not stable across Go releases, and `go run …oapi-codegen@v2.7.0` carried no toolchain pin. CI resolves go1.26.0 from `go.work`; a contributor on a newer Go regenerated a different blob and the gate read it as a missed regeneration. The generated code was byte-identical — the entire diff was one hunk inside `swaggerSpec`.

`OAPI_CODEGEN` now pins `GOTOOLCHAIN`, derived from `go.mod` so it cannot drift, matching what CI resolves — the same remedy already used by `scripts/deadcode.sh` and the root Makefile's `GO_TOOLCHAIN`. Codegen was the last tool invocation left unpinned.

### Onboarding

Skills sync ran before the workspace authz step, which is backwards — syncing creates a component OC authorizes against the role that step creates, so a first-time org 403'd on its first action and was told the skills catalogue had failed. Reordered, with skills waiting for authz to *succeed* rather than merely settle.

Review caught a second bug here: the skills mutation's error is sticky, and the row's status read it before `skillsSkipped`, so after "Continue anyway" the step never completed and **"Go to console" never appeared** — the user stranded on a screen whose only exit they had already taken. `"skipped"` is now a status of its own, distinct from both `done` and `error`: completion accepts it, the row still reports the failure, and the error panel withdraws so it stops offering a choice already made.

`GET /config/status` is new: the two connectivity booleans the onboarding gate needs, with none of `GET /config`'s identity or key detail, so the gate works before a caller's permissions are provisioned.

## Test plan

Everything below was run on this branch, not inferred.

| Gate | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test ./...` (aep-api, **with Postgres** via testcontainers) | pass |
| `services/collab` — `tsc --noEmit` + `pnpm test` | 66 pass |
| `apps/console` — `tsc --noEmit` | clean |
| `apps/console` — `vitest run` | 159 files, 2021 tests pass |
| `make license-check` | pass |
| `make deadcode-check` (Go) | `OK — no dead functions` |
| `make deadcode-ts-check` (knip) | pass |
| `make gen-api-check` | pass, and reproducible across toolchains |
| `make lint` | 31/31 turbo tasks, all 3 Go modules `0 issues.` |

Worth knowing when reading the previous run: `gen-api-check` aborted the job, so the dead-code gates, Lint and License check were **skipped** and have never actually run in CI on this branch. They pass locally; this push is their first real exercise.

The DB-backed component tests self-skip without Docker, so they were run **against a real Postgres** rather than left skipped. That run caught two tests that pass locally and would only have failed in CI (`F1`/`F2` used `idp` as a convenient section) — retargeted in their own commit.

### Proof the tests would catch a regression

Every behavioural fix here was verified by reverting it and confirming the matching test fails, then restoring:

| Reverted | Failing tests |
|---|---|
| `connectionConfig.readOnly = !canWrite` | `a viewer's socket is marked read-only…` |
| writer-only `lastToken` | `a viewer's token never becomes the room's commit credential` |
| committer's writer-token selection | `a viewer-triggered flush commits as the room's writer…`, `…with no writer on record does not apply` |
| the `CreateTurn` OR | `ae:resource-config does NOT satisfy it on a real project`, `an unreadable request denies…`, plus `TestChatOperationsCoverTheirRows` |
| skills error read before the skip | `completes after skipping a failed skills sync`, `still shows the skills row as failed after skipping` |
| the onboarding phase order | 4 of that file's cases |

Two of these existed because a test was too weak, not absent. The skip case asserted the button *rendered* and never clicked it, which is how a dead end survived; it now clicks through and asserts `onComplete` fires.

### Verified by extraction, not by eye

The comment-density pass rewrote `permission_gate.go` (505 → 352 lines) and `oc_permissions_catalog.go` (176 → 115). Both tables were extracted and diffed against `HEAD` to prove the change was comment-only: **all 88 gate entries and every OC action list byte-identical.**

## Documentation

`ADR-0027` had drifted into describing a system that no longer exists — 32 gated operations (65), seven permission keys (14), one role (two), `ApplyFiles` ungated (gated), the console untouched (it reads the real scope), and a write/view OR as the intended design (removed). Rewritten to the shipped end state. `internal/authz/README.md` had the same drift and is corrected.

## Known gaps, deliberately

- Three carved-out operations are reached only by the SRE agent's MCP tools; what "authorized" means for an agent rather than a person is undecided. The rest have no caller at all and carry `TODO(authz)`.
- No user-provisioning path enrols a real org member into `ae-admin` or `ae-developer` at runtime. Local dev seeds one fixed account.
- `idp` has no permission, by choice — one arrives with the surface that writes it.


services/aep-api/internal/edge/permission_gate.go
```diff
@@ -0,0 +1,491 @@
...
+	// permission (retired): AppLayout mounts it for the project spec chat
+	// (gated on ae:design — the panel is a write surface, so ae:design-view
+	// alone does not satisfy it there, same read/write split as every other
+	// permission pair in this console) and RegisterFormPage mounts it for
+	// the marketplace registration-form assistant, pointed at a separate
+	// MARKETPLACE_CHAT_PROJECT pseudo-project (gated on ae:resource-config,
+	// its own page's existing permission — nothing to do with design). All
+	// seven gate on the same OR pair (no read/write split within either
+	// context) since the panel is one feature either way.
+	"CreateTurn":         {authz.PermissionDesign, authz.PermissionResourceConfig},
```
kanushka: An `ae:resource-config` caller can still submit `/design` through `CreateTurn` with `collab:false`. Please require `ae:design` for real-project turns and allow resource-config only for the marketplace pseudo-project.

services/aep-api/internal/edge/permission_gate.go
```diff
@@ -0,0 +1,491 @@
...
+	//
+	// ApplyFiles (the room's git-commit flush) needs the stronger ae:design,
+	// exact-match like ProvideDependencyContract/AcceptDependencyAssumption
+	// above and GenerateDesign below — viewing the live room must not imply
+	// permission to commit into it. Deliberately NOT OR'd with
+	// ae:resource-config the way CreateTurn/StreamTurn's shared chat panel
+	// is: the spec collab room is design-track content, and resource
+	// registration (a separate feature/route) has no legitimate reason to
+	// write here.
+	"ValidateCollabAccess": {authz.PermissionDesignView},
```
kanushka, line 157: Please enforce read-only collab connections server-side. UI locks do not stop Yjs updates, and a later editor flush can commit a viewer's changes using the editor's token.

apps/console/src/features/onboarding/components/RepositorySetupStep.tsx
```diff
@@ -0,0 +1,247 @@
...
+    }, 0);
+    return () => clearTimeout(t);
+  }, [phase1Done, authzMutate]);
+
+  function retryAuthz() {
+    setAuthzAttempts((n) => n + 1);
+    authz.mutate();
+  }
+
+  const phase1Status: RowStatus = sync.isError
```
kanushka: After 'Continue anyway', `sync.isError` still takes precedence over `skillsSkipped`, so 'Go to console' never appears even when workspace setup succeeds. Please check `phase1Done` first.

services/aep-api/internal/edge/permission_gate.go
senithkay: Confirmed, and the bypass was real. `TurnInputBody.instruction` carries `/<skill>` flow commands verbatim for the server to expand, and `CreateTurn` forwards the field with no filtering, so `ae:resource-config` alone reached `/design` on a real project — `generate-design`'s exact-match gate only ever protected the console's button.

Fixed in 2341127d, along the lines you suggested. The seven chat rows now declare `ae:design`, and the marketplace pseudo-project is resolved per request against `spec.MarketplaceRegisterProjectID` rather than through an OR. The underlying lesson is in the file header as a rule now: an OR is the wrong tool when two features differ by **path** rather than by operation, because it grants each one's permission everywhere the other reaches.

A request shape the resolver doesn't recognise falls through to `ae:design`, not to the marketplace permission — an omission there costs a marketplace caller a visible 403, where the other direction would hand a real project's design controls to a permission that shouldn't reach them.

Tests cover all four combinations per operation, plus `TestChatOperationsCoverTheirRows` so the project-scoped set and the rows can't drift apart. Restoring the OR fails both the behavioural test and that consistency check.

services/aep-api/internal/edge/permission_gate.go
senithkay: Agreed on both halves, and this is fixed in 7ea8ff42 (pushed before this review landed, so it may not have been visible when you commented).

`validate-collab-access` now also answers `canWrite` — whether the joiner holds `ae:design`, as opposed to the `ae:design-view` that merely admits them — and `services/collab` acts on it twice:

- `onAuthenticate` sets `connectionConfig.readOnly`, so Hocuspocus drops that socket's `messageYjsUpdate` and `syncStep2` outright. Verified against the installed v4.3.0 source rather than assumed.
- The participant's token is kept out of the commit path and they're left off the `Co-authored-by` trailers.

The second half turned out to matter for a reason beyond permissions, which is the part I'd flag: the room's `lastToken` authenticates the unload flush and was simply the most recent joiner's. Once `ApplyFiles` started requiring `ae:design`, a single viewer arriving in a room would have become the credential every later flush used — failing all of them and stranding the actual authors' uncommitted work until someone with write access reconnected. A viewer can still *trigger* a flush; it now commits as the room's last writer, and is skipped entirely when no writer is on record rather than attempted with credentials that cannot work.

An older BFF that omits `canWrite` reads as read-only: a missing answer about permission shouldn't grant one.

Each of the three changes was verified by reverting it and confirming the matching test fails — 3 of 3 caught.

apps/console/src/features/onboarding/components/RepositorySetupStep.tsx
senithkay: Good catch — confirmed by reproducing it as a failing test before touching the code. The user was stranded on a screen whose only exit they'd already taken.

Fixed in 78ba666d. I went slightly past "check `phase1Done` first", because ordering alone would have reported a skip as success: `skillsDone` is `sync.isSuccess || skillsSkipped`, so reading it before the error makes a skipped step render as a green tick and the closing copy claim the skills landed.

Instead `"skipped"` is now its own `RowStatus`, distinct from both `done` and `error` — the step didn't succeed, the user chose to proceed, and the org is usable without it. Completion accepts `done` **or** `skipped` for the skills row while the workspace row must still be genuinely `done` (it's the hard gate), and the skills error panel is withdrawn once skipped so it stops offering a choice already made next to the button that acts on it.

Worth noting this predates the phase reorder in this PR, which preserved it — and it survived my own test because that test asserted the button *rendered* and never clicked it. The new cases click through and assert `onComplete` fires; reverting the fix fails two of them.

PR #788, opened 2026-09-17
Run the generated app locally, and let it start where no gateway assertion exists

xlight05:
Seven commits. The first six make a generated app runnable on a laptop; the last
lets one start in an environment whose gateway was provisioned before assertions
existed.

## `wire` — run what was built

`pnpm play <dir> wire` stands the generated app up locally: compose for the
backend, the dev server as the gateway. Standing in for the gateway is exactly
holding the private key whose certificate the service was given, so `wire` mints
a keypair per project and signs an assertion per role. ADR-0002 records the
shape. An app whose `mock/` predates wired mode is refused rather than half-run.

Alongside it: the skill fixes found by actually running generated apps, a runner
fix (the task-list tools have to be named to be present — not-denied is not the
same as present), and the wired-mode review's findings.

## Starting without a gateway assertion

Every environment gateway provisioned before 2026-09-16 carries only
`jwtauth_v1`. `backendjwt_v1` landed in `setup-environment-gateway.sh` that day
(8a9a0517), so those gateways mint no `x-jwt-assertion`, the three
`GATEWAY_ASSERTION_*` variables are never set, and both assets panicked at init.
A generated service in such an environment has never started — confirmed on the
ae-demo cluster, where no service in any `dp-default-*` namespace has ever had
those variables.

Both assets gain a second branch:

| Variables | Behaviour |
|---|---|
| all three set | unchanged — the assertion is verified as before |
| all three absent | the caller is read from `x-forwarded-authorization` and **decoded without any signature check** |
| some set | still fatal — a half-configured assertion is a broken deployment, not an unprovisioned one |

**The middle row gives away the service's only trust anchor.** Any pod that can
reach the service can name itself any user and grant itself any scope. The
gateway is then the only thing between the internet and the data, and nothing
inside the cluster is standing there at all. The asset says so in a banner with
the exact `curl`, and the process says so in a warning on every boot.

This is a stop-gap with a known expiry. The fix is to provision the environment
gateway's backend-JWT keypair; when that lands, **delete the fallback, not the
verification.**

Tracked for removal in #789, which also records what blocks it: the
`setup-environment-gateway.sh` BIND guard refuses to upgrade a release the
script itself created, so the keypair cannot currently be rolled out to any
environment that already exists.

Claims absent from the access token are backfilled from the gateway's
claim-mapped `x-user-*` headers, so an IdP that publishes no `username` or
`ouHandle` does not strip the caller of their identity for no visible reason.

The `SKILL.md` files carry none of this — the assets are copied verbatim into
every app and document themselves. The only doc changes are six clause-level
corrections where the text had become false (net zero lines).

### Verification

- Ballerina: the four existing assertion tests pass **unchanged**; five new
  fallback-mode tests (unsigned token taken wholesale, unreadable → 401 not
  anonymous, header backfill, bare token); a partial trio panics.
- Go: `gofmt` and `go vet` clean; five tests including partial-config-still-fatal.
- The real generated `gym-api` package compiles with the new asset.

Not yet proven against a live gateway: that `x-forwarded-authorization` actually
arrives and carries the claims the app needs rests on the P2-live-1/P2-live-2
spike findings, not on a run made for this change.


kanushka:
There are several @coderabbitai review comments as well. Please address those in a separate PR.

kanushka (approved):

PR #797, opened 2026-09-18
Deployments: setup seeds the validation annotation, so a fresh cluster keeps its Validation step

axewilledge:
A cluster built by `setup.sh` comes up with **no Validation step** on its Deployments card. Both scripts that create `Environment/default` now seed the annotation that draws it.

## Why it disappeared

The console builds an environment's steps as a consequence, never a stored number:

```
steps(env) = [ Deployment ]
           + [ Validation ]          if aep.wso2.com/validation == "on"
           + [ Promote to <next> ]   if something follows
```

The BFF normalises the annotation on the way out:

```go
// Absent means off. Anything we do not recognise also means off —
// an unreadable annotation must not switch validation on.
if e.Validation != "on" { e.Validation = "off" }
```

That default is right: a value nobody can read should never turn a gate **on**. But nothing ever wrote the annotation. `setup-aep.sh` and `setup-aectl.sh` both create `Environment/default` with a bare `spec`, so every cluster these scripts build starts without the step, and the only way to get it back is to annotate the object by hand — which a cluster rebuild then throws away.

## The fix

Both scripts now create the environment with:

```yaml
metadata:
  annotations:
    aep.wso2.com/validation: "on"
```

`default` is the entry environment on a local cluster — where AEP builds, deploys and validates — so it is the one that has always had the step. This restores the behaviour that existed before environments became configurable, rather than introducing a new one.

## Verified

`kubectl apply --dry-run=server` against a live cluster returns `environment.openchoreo.dev/default configured` — the manifest parses, the API server accepts it, and `apply`'s merge adds the annotation without disturbing the `aep.wso2.com/thunder-*` annotations the controllers write onto the same object. Both scripts pass `bash -n`.

## What this does NOT fix

**An environment created any other way still starts without the step.** These two scripts cover the local cluster's `default`; anything created through OpenChoreo directly, or by another tool, is unannotated and therefore silently loses a step in the console.

Two things would close that properly, and this PR is deliberately neither:

- The console defaults the **entry** environment to `on` when the annotation is absent, and every other environment to `off`. No platform write, preserves old behaviour everywhere — but hides a default in code.
- Environment creation, wherever it happens, is responsible for the annotation. Honest about where configuration lives, but needs every creation path to know.

That choice belongs in ADR-0033, which is still owed from the multi-environment work.

## A sharp edge worth knowing

The comparison is `!= "on"`, exactly. `"On"`, `"true"`, `"enabled"` or a trailing space all normalise silently to `off`, with no warning anywhere in the stack. Anyone setting this by hand gets no feedback that they got it wrong — the step simply does not appear.


xlight05 (approved):

deployments/scripts/setup-aectl.sh
```diff
@@ -635,6 +635,14 @@ kind: Environment
...
   name: default
   namespace: default
+  annotations:
+    # Whether the console draws a Validation step for this environment. The
+    # console reads it per environment and treats ABSENT as "off", so an
+    # environment nobody annotates silently loses the step — which is what
+    # happened to every cluster built before this line existed. "default" is
+    # the entry environment here: it is where AEP builds, deploys and
+    # validates, so it is the one that has always had the step.
+    aep.wso2.com/validation: "on"
```
kaje94, line 645: would prefer a name like `xxx/validate-deployment: true`

